PR #174, opened 2014-12-21
Analog config

soundanalogous:
This is an evolution of https://github.com/firmata/arduino/pull/40 so I opened a new pull for it. However adding this functionality is not so straight forward. There are 2 major issues:
1. Only SAM-based variants include the methods `analogReadResolution` and `analogWriteResolution`.
2. Only AVR-based variants support setting the analogReference value. SAM variants have an `analogReference` method, but it is set with a typedef value rather than a constant. Furthermore, the DEFAULT and EXTERNAL are defined for SAM variants, but they cannot be passed to `analogReference` since `analogReference` expects a parameter of type `eAnalogReference` which is a type created specifically for SAM variants.

There appears to be no clean implementation possible here. I'd have to pick an existing define that is unique to each architecture and use that to determine whether a board is of SAM or AVR architecture, or I'd have to add some defines to Boards.h to sort it out.

### DIFF
--- a/examples/StandardFirmata/StandardFirmata.ino
+++ b/examples/StandardFirmata/StandardFirmata.ino
@@ -84,12 +84,7 @@ unsigned int i2cReadDelayTime = 0;  // default delay time between i2c read reque
 
 byte analogReadRes = 10; // default - better to define in Boards.h?
 byte analogWriteRes = 8; // default - better to define in Boards.h?
-#ifdef DEFAULT
-// DEFAULT is not always = 0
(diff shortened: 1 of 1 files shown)


soundanalogous:
Teensy 3.1 can also make use of analogWriteResolution

soundanalogous:
Arduino Zero uses 12-bit ADC

soundanalogous:
ESP8266 has a 10-bit PWM resolution

PR #176, opened 2014-12-21
replace error strings with error codes

soundanalogous:
This update frees 122 bytes of RAM. I have replace the old string error messages in StandardFirmata with single byte error codes (0 - 127). A user can add custom error codes to any Firmata sketch. The client library should provide a message for each error code (or otherwise handle the error as necessary).

This update may impact any client library that handles the old string error messages sent by StandardFirmata (if this is actually the case with any client library - I suspect it is rare). However, given that this pull request is for Firmata 2.4.0 and the RAM savings and overall more explicit nature of the new error_code I think this update justifies the backwards incompatible change.

See conversation here: https://github.com/firmata/arduino/pull/80


soundanalogous:
Closing this since it is not likely to be the implementation. Still planning to revisit for Firmata 3.0

PR #208, opened 2015-07-26
Point to correct library path in Arduino.app

beriberikix:
Libraries are stored in a different path in the latest Arduino.app

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -107,8 +107,8 @@ You will first need to remove the existing Firmata library, then clone firmata/a
 into an empty Firmata directory:
 
 ```bash
-$ rm -r /Applications/Arduino.app/Contents/Resources/Java/libraries/Firmata
-$ git clone git@github.com:firmata/arduino.git /Applications/Arduino.app/Contents/Resources/Java/libraries/Firmata
+$ rm -r /Applications/Arduino.app/Contents/Java/libraries/Firmata/
+$ git clone git@github.com:firmata/arduino.git /Applications/Arduino.app/Contents/Java/libraries/Firmata/
 ```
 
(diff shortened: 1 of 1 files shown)


soundanalogous:
I need to update the installation instructions. As of Arduino 1.6.4 Firmata can simply be updated via the Arduino Library Manager:

In the IDE:
1. navigate to: Sketch -> Include Library -> Manage Libraries
2. Filter "Firmata"
3. Tap on the Firmata list item
4. Select the version you want to install (or switch back to) and click Install

This only works for what is in the Master branch. For other branches you'd need to copy Firmata into the 3rd party libraries folder (this works now, but maybe only after using the library manager at least once I think) rather than having to copy into the application directory.

beriberikix:
Ah, right, so glad they did that. But what if you're building from src? For example, I need to modify Boards.h, which could theoretically get massive. It make more sense for the edge-case users to use a forked build, no?

soundanalogous:
Yeah but you should be able to simply clone Firmata to ~/Documents/Arduino/libraries now. That is independent from the Arduino IDE so when you update the IDE you don't need to re-clone.

soundanalogous:
However I have not tried this yet by cloning to libraries dir. It may conflict with the arduino library manager which is targeting the same directory.

beriberikix:
Oh, duh.

beriberikix:
FWIW, I did a fresh install of Arduino on OSX and cloning into /libraries works like a charm. :+1: 

screen shot 2015-08-02 at 10 49 42 am

soundanalogous:
Cool I'll update the instructions in the readme soon.

PR #210, opened 2015-08-08
Chip kit support

EmbeddedMan:
Adding chipKIT board support to Firmata.

### DIFF
--- a/examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
+++ b/examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
@@ -1,19 +1,17 @@
 /*
- * Firmata is a generic protocol for communicating with microcontrollers
- * from software on a host computer. It is intended to work with
- * any host computer software package.
- *
- * To download a host software package, please clink on the following link
- * to open the download page in your default browser.
- *
- * http://firmata.org/wiki/Download
- */
+  Firmata is a generic protocol for communicating with microcontrollers
+  from software on a host computer. It is intended to work with
+  any host computer software package.
 
-/*
+  To download a host software package, please clink on the following link
+  to open the download page in your default browser.
+
+  https://github.com/firmata/arduino#firmata-client-libraries
+
   Copyright (C) 2006-2008 Hans-Christoph Steiner.  All rights reserved.
   Copyright (C) 2010-2011 Paul Stoffregen.  All rights reserved.
   Copyright (C) 2009 Shigeru Kobayashi.  All rights reserved.
-  Copyright (C) 2009-2014 Jeff Hoefs.  All rights reserved.
+  Copyright (C) 2009-2015 Jeff Hoefs.  All rights reserved.
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
@@ -22,29 +20,25 @@
 
   See file LICENSE.txt for further informations on licensing terms.
 
-  formatted using the GNU C formatting and indenting
+  Last updated by Jeff Hoefs: April 11, 2015
 */
-
-/*
- * TODO: use Program Control to load stored profiles from EEPROM
- */
 
 #include <SoftPWMServo.h>  // Gives us PWM and Servo on every pin
 #include <Wire.h>
 #include <Firmata.h>
 
-// move the following defines to Firmata.h?
-#define I2C_WRITE B00000000
-#define I2C_READ B00001000
-#define I2C_READ_CONTINUOUSLY B00010000
-#define I2C_STOP_READING B00011000
-#define I2C_READ_WRITE_MODE_MASK B00011000
+#define I2C_WRITE                   B00000000
+#define I2C_READ                    B00001000
+#define I2C_READ_CONTINUOUSLY       B00010000
+#define I2C_STOP_READING            B00011000
+#define I2C_READ_WRITE_MODE_MASK    B00011000
 #define I2C_10BIT_ADDRESS_MODE_MASK B00100000
+#define MAX_QUERIES                 8
+#define REGISTER_NOT_SPECIFIED      -1
 
-#define MAX_QUERIES 8
+// the minimum interval for sampling analog input
 #define MINIMUM_SAMPLING_INTERVAL 10
 
-#define REGISTER_NOT_SPECIFIED -1
 
 /*==============================================================================
  * GLOBAL VARIABLES
@@ -65,7 +59,7 @@ int pinState[TOTAL_PINS];           // any value that has been written
 /* timer variables */
 unsigned long currentMillis;        // store the current value from millis()
 unsigned long previousMillis;       // for comparison with currentMillis
-unsigned int samplingInterval = 19;          // how often to run the main loop (in ms)
+unsigned int samplingInterval = 19; // how often to run the main loop (in ms)
 
 /* i2c data */
 struct i2c_device_info {
@@ -77,12 +71,11 @@ struct i2c_device_info {
 /* for i2c read continuous more */
 i2c_device_info query[MAX_QUERIES];
 
-boolean isResetting = false;
-
 byte i2cRxData[32];
 boolean isI2CEnabled = false;
 signed char queryIndex = -1;
-unsigned int i2cReadDelayTime = 0;  // default delay time between i2c read request and Wire.requestFrom()
+// default delay time between i2c read request and Wire.requestFrom()
+unsigned int i2cReadDelayTime = 0;
 
 SoftServo servos[MAX_SERVOS];
 byte servoPinMap[TOTAL_PINS];
@@ -90,6 +83,26 @@ byte detachedServos[MAX_SERVOS];
 byte detachedServoCount = 0;
 byte servoCount = 0;
 
+boolean isResetting = false;
+
+/* utility functions */
+void wireWrite(byte data)
+{
+#if ARDUINO >= 100
+  Wire.write((byte)data);
+#else
+  Wire.send(data);
+#endif
+}
+
+byte wireRead(void)
+{
+#if ARDUINO >= 100
+  return Wire.read();
+#else
+  return Wire.receive();
+#endif
+}
 
 /*==============================================================================
  * FUNCTIONS
@@ -139,11 +152,7 @@ void readAndReportData(byte address, int theRegister, byte numBytes) {
   // do not always require the register read so upon interrupt you call Wire.requestFrom()
   if (theRegister != REGISTER_NOT_SPECIFIED) {
     Wire.beginTransmission(address);
-#if ARDUINO >= 100
-    Wire.write((byte)theRegister);
-#else
-    Wire.send((byte)theRegister);
-#endif
+    wireWrite((byte)theRegister);
     Wire.endTransmission();
     // do not set a value of 0
     if (i2cReadDelayTime > 0) {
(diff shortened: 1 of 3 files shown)


examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
```diff
@@ -0,0 +1,735 @@
...
+  if (pin < TOTAL_PINS) {
+    switch (pinConfig[pin]) {
+      case SERVO:
+        if (IS_PIN_DIGITAL(pin))
+          servos[servoPinMap[pin]].write(value);
+        pinState[pin] = value;
+        break;
+      case PWM:
+        if (IS_PIN_PWM(pin))
+          servoWrite(PIN_TO_PWM(pin), value);
```
soundanalogous: Without SoftPWMServo.h that library does the chipKIT have no PWM functionality?

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
```diff
@@ -0,0 +1,735 @@
...
+  // TODO: save status to EEPROM here, if changed
+}
+
+void analogWriteCallback(byte pin, int value)
+{
+  if (pin < TOTAL_PINS) {
+    switch (pinConfig[pin]) {
+      case SERVO:
+        if (IS_PIN_DIGITAL(pin))
+          servos[servoPinMap[pin]].write(value);
```
soundanalogous: This will break if anyone tries to write to a servo using any existing Firmata client library that supports servos. You will need to move servoWrite here and handle and PWM conversion as necessary or it will not be compatible with existing Firmata client libs.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
EmbeddedMan: No, we have our own version of Servo.h that works just fine on some pins. But SoftPWMServo works on every pin, thus it's a better choice for Firmata, I feel.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
```diff
@@ -0,0 +1,735 @@
...
+        pinConfig[pin] = PWM;
+      }
+      break;
+    case SERVO:
+      if (IS_PIN_DIGITAL(pin)) {
+        pinConfig[pin] = SERVO;
+        if (servoPinMap[pin] == 255 || !servos[servoPinMap[pin]].attached()) {
+          // pass -1 for min and max pulse values to use default values set
+          // by Servo library
+          attachServo(pin, -1, -1);
```
soundanalogous: If SoftPWMServo.h doesn't have an `attach` method, this will break for pretty much all up-to-date Firmata client libraries with Servo support. It looks like adding servo support for chipKIT may actually require a lot more changes.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
EmbeddedMan: And our version of analog output (PWM output) also works just fine, but again can only be done with a small number of pins. SoftPWMServo allows us to do PWM (Analog output) and Servo on all pins.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
soundanalogous: One solution may be to only set the pinConfig here and remove attachServo and detachServo (assuming there is not equivalent in SoftPWMServo.h).

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
EmbeddedMan: Hmm. I guess I'm not quite following. Any pin that can be a digital I/O on
PIC32 can be PWM or Servo outputs as well (using SoftPWMServo library) so
there shouldn't be any reason for the client libraries to worry about what
pins they use for PWM or Servo outputs. Or am I missing the point?

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
```diff
@@ -0,0 +1,735 @@
...
+      break;
+    case SERVO_CONFIG:
+      if (argc > 4) {
+        // these vars are here for clarity, they'll optimized away by the compiler
+        byte pin = argv[0];
+        int minPulse = argv[1] + (argv[2] << 7);
+        int maxPulse = argv[3] + (argv[4] << 7);
+
+        if (IS_PIN_DIGITAL(pin)) {
+          if (servoPinMap[pin] < MAX_SERVOS && servos[servoPinMap[pin]].attached()) {
```
soundanalogous: You may need to remove lines 520 - 523. Allow SERVO_CONFIG to set the pin mode. I'm not sure if SoftPWMServo.h has an equivalent of setting a min and max pulse. If so it should be set here.

soundanalogous:
It looks like this may need more work. You should test servo support against a popular Firmata client library like (johnny-five)[https://github.com/rwaldron/johnny-five]. I know @rwaldron is eager to get chipKIT support for johnny-five so maybe he is also available to do some testing here. I don't have any chipKIT boards to test with myself so I'll be relying on you to ensure it's working well with popular client libraries.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
soundanalogous: Ok this will work fine then.

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
soundanalogous: With most Firmata client libraries, when someone configures a pin for Servo, that pin will be marked as a Servo type and line 323 will be called when the user writes a value to that servo, not line 290. Sure anyone could configure a pin as PWM and write to it but that will be unexpected for many existing client library users. They would have to do things differently when using a chipKIT board vs an Arduino board for example.

soundanalogous: Edit in previous comment... line 328 instead of line 290.

soundanalogous: I think all you may need to do here actually is replace line 323 with: `servoWrite(PIN_TO_PWM(pin), value);`

That way it will work whether people use a pin configured as Servo or a pin configured as PWM (they are different configurations in Firmata).

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
soundanalogous: I should have looked at this first: https://github.com/chipKIT32/chipKIT32-MAX/blob/master/hardware/pic32/libraries/SoftPWMServo/SoftPWMServo.h

EmbeddedMan:
Excellent suggestion. I've been going through every johnny-five Servo based example (that I have hardware for - i.e. not the I2C based expander ones) and everything works great, with one minor exception. The only API call I can find that doesn't seem to do what it's supposed to is servo.stop(). I'll figure out why that's not working, and then update the PR. Every other servo API in johnny-five seems to work identically on chipKIT and Arduino boards.

soundanalogous:
@EmbeddedMan is it possible to use an Arduino ethernet shield with a chipKIT board?

examples/StandardFirmataChipKIT/StandardFirmataChipKIT.ino
```diff
@@ -0,0 +1,735 @@
...
+  Copyright (C) 2009-2015 Jeff Hoefs.  All rights reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  See file LICENSE.txt for further informations on licensing terms.
+
+  Last updated by Jeff Hoefs: April 11, 2015
```
soundanalogous: Add your name and the date here

rwaldron:
>  The only API call I can find that doesn't seem to do what it's supposed to is servo.stop(). I'll figure out why that's not working,

Keep in mind that `stop` is meaningless unless the program: 

- has specified a time-to-position for standard servos
- is controlling a continuous servo and the _correct_ neutral point is known (generally, that defaults to 90°, but some continuous servos don't have manual calibration, but instead allow a wide neutral zone)

EmbeddedMan:
Ohhh! In that case, stop is working great. :-) I thought servo.stop();
would kill the pulses coming out of the pin (i.e. "stop" sending RC servo
pulses). But you're saying that it actually only stops the pulses from
changing (for standard servos) or moves the pulses back to neutral point
(probably 1.5ms) for continuous rotation.

Let me do some more testing, and I'll let you know if the everything is
working correctly as far as stop. (I want to confirm my understanding based
on your description of what stop does.)

EmbeddedMan:
OK, I've done more testing comparing my PR chipKIT version of StandardFirmata with the normal Arduino version, with respect to the Servo API. I can find no differences. Stop() appears to work exactly the same on both boards. I've also updated StandardFirmataChipKIT.ino with my name.

Are there any other things you'd like to see fixed before moving forward with the PR?

(Oh, and I believe that there are some folks who have gotten the Arduino Ethernet shield to work with a chipKIT board, but I've never tried it, and it sounds like it takes some hardware modifications to the shield to get it to work.)

soundanalogous:
I think this is good to go then. I'll merge it and then make some minor changes as I want to pull in some changes I made in the v2.5beta1 branch (basically everything except the addition of SET_DIGITAL_PIN_VALUE which requires the bump from 2.4 to 2.5). After I make those changes I can tag a new release (Firmata 2.4.4) and it will show up for users in the Arduino IDE library manager.

soundanalogous:
Who should I credit in the release notes for the contribution? Yourself and Rick Anderson? I'm not sure what the extend of the collaboration is on the ChipKIT side.

EmbeddedMan:
Jeff,

Thanks for asking. It should be me, Rick Anderson and Keith Vogel.

*Brian

soundanalogous:
How soon will ChipKIT support be available in the Arduino Board manager?

EmbeddedMan:
Jeff, we're not sure. It's one of the things Rick is currently working on,
and we're getting close, but we're not there yet. Maybe a few weeks?

*Brian

soundanalogous:
Okay, so until then you just merge Firmata master into MPIDE?

EmbeddedMan:
Yup, I just took care of putting my latest files in there, and when you do
your new release, I'll update them again with your latest.

soundanalogous:
Done! https://github.com/firmata/arduino/releases/tag/v2.4.4

Let me know if I should change anything in the first couple of sentences regarding how people can start using Firmata with ChipKIT.

EmbeddedMan:
Jeff, that looks fantastic. My only suggestion would be to say something
like "Be sure to use MPIDE 20150801 or later to get the most out of
Firmata" or something along those lines. (It wasn't until 20150801 that we
made the necessary changes in our own variant files to properly handle
analog inputs with Firmata on every board.)

*Brian

soundanalogous:
Looks like the default download for OS X is 20140821.

soundanalogous:
Should the instructions say to download one of the 20150802 test releases from the top of this list?
http://chipkit.s3.amazonaws.com/index.html

EmbeddedMan:
Jeff,

Well, users should probably download the latest test release, whatever
version that is, but it should be at least 20150801. The problem with
showing that list is that it has all of our builds in it, and it can be
confusing for somebody who's not familiar with the current state of chipKIT
development. For example, you probably don't want one of the 0023 builds,
but should rather start with the 0150 builds (the prefix at the beginning
of the name). So, at this point in time, the right build for folks is the
builds/mide-0150-XXXXXXX-20150801-test.zip builds.

*Brian

EmbeddedMan:
And yes, the state of chipKIT releases - at this point in time - is not
only kind of a mess but is not documented properly. We will be fixing that
in the near future, when we have chipKIT-core working 100% and we can point
all users to that as being the 'latest' (from within Arduino IDE of course).

*Brian

rwaldron:
> But you're saying that it actually only stops the pulses from
changing (for standard servos) or moves the pulses back to neutral point
(probably 1.5ms) for continuous rotation.

You got it :)

PR #220, opened 2015-10-05
Adding Pinoccio Scout to Boards.h

makenai:
This adds a definition to Boards.h for the Pinoccio Scout, an Arduino compatible device similar to the Mega, as seen here: https://pinocc.io/

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -565,17 +565,15 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define TOTAL_ANALOG_PINS       8
 #define TOTAL_PINS              NUM_DIGITAL_PINS // 32
 #define VERSION_BLINK_PIN       23
-#define PIN_SERIAL1_RX          0
-#define PIN_SERIAL1_TX          1
-#define PIN_SERIAL2_RX          13
-#define PIN_SERIAL2_TX          14
+#define PIN_SERIAL1_RX          13
+#define PIN_SERIAL1_TX          14
 #define IS_PIN_DIGITAL(p)       (((p) >= 2) && ((p) <= 8)) || (((p) >= 24) && ((p) <= 31))
 #define IS_PIN_ANALOG(p)        ((p) >= 24 && (p) <= 31)
 #define IS_PIN_PWM(p)           digitalPinHasPWM(p)
 #define IS_PIN_SERVO(p)         IS_PIN_DIGITAL(p)
 #define IS_PIN_I2C(p)           ((p) == SCL || (p) == SDA)
 #define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
-#define IS_PIN_SERIAL(p)        ((p) == 0 || (p) == 1 || (p) == 13 || (p) == 14)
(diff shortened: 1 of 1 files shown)


soundanalogous:
Please also add the UART pins. [See how I've done this for Mega](https://github.com/firmata/arduino/blob/master/Boards.h#L203-L208). It should be similar for Pinoccio.

Also add the `IS_PIN_SERIAL` macro. Looks like this for Mega: https://github.com/firmata/arduino/blob/master/Boards.h#L215.

makenai:
@soundanalogous Thanks - I've rebased and added the UART pins and macro per your recommendation

Boards.h
```diff
@@ -560,6 +560,26 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+// Pinoccio Scout
+#elif defined(ARDUINO_PINOCCIO)
+#define TOTAL_ANALOG_PINS       8
+#define TOTAL_PINS              NUM_DIGITAL_PINS // 32
+#define VERSION_BLINK_PIN       23
+#define PIN_SERIAL1_RX          0
+#define PIN_SERIAL1_TX          1
+#define PIN_SERIAL2_RX          13
+#define PIN_SERIAL2_TX          14
+#define IS_PIN_DIGITAL(p)       (((p) >= 2) && ((p) <= 8)) || (((p) >= 24) && ((p) <= 31))
```
soundanalogous: Do pins 0 and 1 map to Serial or Serial1? If they map to Serial1, the you should be able to use pins 0 and 1 as digital inputs unless those pins don't have digital I/O for some other reason.

soundanalogous: Okay I found the pinout for the Pinoccio. They did tie the usb serial to digital pins 0 and 1. The number at the end of PIN_SERIAL aligns with the Serial object, so PIN_SERIAL1 is for Serial1, PIN_SERIAL2 is for Serial2. You should remove lines 568 and 569 and change the following 2 lines to PIN_SERIAL1 (so you'll only map to pins 13 and 14 for now since the default Serial object is reserved for Firmata communication). At some point I'll open it up when using Firmata via a non-serial channel such as Ethernet or WiFi.

makenai:
Thanks for looking that up for me and for your patience in explaining it. I was wondering why the numbering started at 1 and not 0, so that makes perfect sense.

Per your earlier comment, I've also extended the range of digital pins to include 9-16 which are mapped this way:

```c
#define SS              9
#define MOSI            10
#define MISO            11
#define SCK             12
#define RX1             13
#define TX1             14
#define SCL             15
#define SDA             16
```

Does that look correct?

soundanalogous:
If those pins can be used as digital I/O pins then they should be included in the range of digital pins. You should test it against StandardFirmata if you have not already. Users may not know which digital pin numbers to specify for those pins since they fall out of order as they are arranged on the board. It would be helpful to document this somewhere. Add a comment to the code here. It may end up being a bit of a hidden feature.

makenai:
Roger that. I've verified all the digital pins work to blink an LED and added a comment to explain the numbering.

soundanalogous:
Cool I think we're good here. Merging...
Thanks!

Issue #82, opened 2013-08-17
Call Wire.endTransmission but keep the connection active with Wire.endTransmission(false);?


rwaldron:
https://github.com/firmata/arduino/blob/master/examples/StandardFirmata/StandardFirmata.ino#L101

Working on reading a MMA8452Q (triple axis accelerometer), ran the sketch [here](https://github.com/sparkfun/MMA8452_Accelerometer) and it works fine, but no dice over Firmata. Changing [this line](https://github.com/firmata/arduino/blob/master/examples/StandardFirmata/StandardFirmata.ino#L101) to `Wire.endTransmission(false);` (as seen in the sketch) makes it work correctly.

I'm wondering what, if any, implications exist by doing this?

MrYsLab:
I think there are some errors in the i2c-restart branch. I tested the code and I had to do the following to make it work:

Original:
`#define I2C_END_TX_MASK               B01000000
#define I2C_STOP_TX                   1
#define I2C_RESTART_TX                0`

Changed to:
`#define I2C_END_TX_MASK               B01000001
#define I2C_STOP_TX                   1
#define I2C_RESTART_TX                0`

Original:
`if (argv[1] & I2C_END_TX_MASK) {
      stopTX = I2C_RESTART_TX;
    }
    else {
      stopTX = I2C_STOP_TX; // default
    }`

Changed to:
` if (argv[1] & I2C_END_TX_MASK) {
      stopTX = I2C_STOP_TX;
    }
    else {
      stopTX = I2C_RESTART_TX; // default
    }`

soundanalogous:
You should not need to change the mask since the lower 3 bits are always 0 (they are reserved for 10-bit address mode). If you're seeing any of the lower 3 bits of argv[1] set, double check how you are packing bits before sending on the python client side.

The logic is tricky regarding stop and restart since I had to invert it to maintain backwards compatibility. True is false and false is true. So if bit 6 in argv[1] is set, that means endTransmission should be passed false and vice versa. The mask should force all other bits to zero for the comparison so the condition will be true if bit 6 is set and false if it is not set.

soundanalogous:
Not sure if [line 450](https://github.com/firmata/arduino/compare/v2.5.beta1...i2c-restart) needs to be false as well. I think I left it out when I originally wrote this patch since it was only called once and therefore not likely needing to restart. Something to try however if this patch doesn't seem to be working. I still don't have any i2c devices with this behavior to test with myself.

[...]

PR #226, opened 2015-10-11
enable option to auto restart i2c transmission per issue #82

soundanalogous:
see #82


soundanalogous:
According to #82 and https://github.com/rwaldron/johnny-five/issues/322#issuecomment-163441035 this patch works.

PR #227, opened 2015-10-11
Add support for Arduino Zero

soundanalogous:
You will need an official Arduino Pro Zero board. Do not get an M0 Pro board, that will not work with Firmata.

In order to use, you'll have to make sure the Servo library is at least version 1.1.0.  Follow these instructions:
1. In the Arduino IDE (the arduino.cc one), go to Sketch > Include libraries > Manage libraries.
2. Search for "Servo" and scroll down the results list to "Servo by Michael Margolis..."
3. Tap on the list item (I wish this was more clear in the UI), then an option to Update will appear
4. Click on update.

TODO:
- [x] Ensure only digital I/O pins are reset in `systemResetCallback`
- [ ] Thorough testing

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -242,6 +242,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         ((p) - 2)
 
 
+// Arduino Zero
+// Note this will work with an Arduino Zero Pro, but not with an Arduino M0 Pro
+// Arduino M0 Pro does not properly map pins to the board labeled pin numbers
+#elif defined(_VARIANT_ARDUINO_ZERO_)
+#define TOTAL_ANALOG_PINS       6
+#define TOTAL_PINS              25 // 14 digital + 6 analog + 2 i2c + 3 spi
+#define TOTAL_PORTS             3 // set when TOTAL_PINS > num digitial I/O pins
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define DAC_RESOLUTION          10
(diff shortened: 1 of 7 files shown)


soundanalogous:
This is going to push out the v2.5 release at least a week while I wait for my Zero board to arrive.

rwaldron:
Can you link to the Zero you ordered? I don't want to order the wrong thing again.

soundanalogous:
I ordered direct from arduino.cc: http://store-usa.arduino.cc/products/abx00003

soundanalogous:
Zero does not yet seem to be very stable yet. I'm seeing strange issues like digital pins no longer working when a pwm pin is in use on the same port. Need to do more testing with regular Arduino sketches to isolate the issues (what is Firmata-related and what is weirdness with Zero).

soundanalogous:
Zero does not like this: https://github.com/firmata/arduino/blob/master/examples/StandardFirmata/StandardFirmata.ino#L266. Commenting out that line will enable digital input. However it still does not solve all problems I'm seeing.

soundanalogous:
Issue with Zero per previous comment is describe here: https://github.com/firmata/arduino/issues/231

soundanalogous:
Turns out for Zero, internal pull-ups can only be set by setting the pinMode to INPUT_PULLUP so the last commit does not help the Zero situation.

soundanalogous:
Fixed internal pull-up issue in [last commit](https://github.com/firmata/arduino/commit/8ab43c3f5be036f87da27b64a13708c6895c653f).

PR #246, opened 2015-12-05
dht21/22

darashenka:
dht21/dht22 for firmata

### DIFF
--- a/examples/DHT21Firmata/DHT21Feature.cpp
+++ b/examples/DHT21Firmata/DHT21Feature.cpp
@@ -1,0 +1,75 @@
+/*
+ * FILE: DHT21Firmata.cpp
+ * VERSION: 0.1
+ * PURPOSE: DHT21 Temperature & Humidity Sensor feature library for Firmata
+ * LICENSE: GPL v3 (http://www.gnu.org/licenses/gpl.html)
+ *
+ * to be used with Firmata
+ * URL: http://www.firmata.org
+ * uses DHT21lib
+ * URL: http://arduino.cc/playground/Main/DHT21Lib
+ *
+ * HISTORY:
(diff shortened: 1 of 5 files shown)


soundanalogous:
ConfigurableFirmata has it's own repo now: https://github.com/firmata/ConfigurableFirmata

Also the way I'd like to add DHT21/22 support is how Encoder was added, as a separate repo. See FirmataEncoder for an example: https://github.com/firmata/FirmataEncoder.

Users can then add DHT21 support using [http://firmatabuilder.com].

soundanalogous:
Alternatively this could be a good case for a DHT humidity sensor backpack due to the timing requirements that I've seen in various implementation examples. Seems there are some delays that cannot be avoided and an i2c backpack would be a way to get around it.

https://github.com/ajfisher/nodebots-interchange.

darashenka:
Hey, Jeff!

nice, thanks!

I'll fix in https://github.com/firmata/ConfigurableFirmata. 
I would like to have a common-confiurable instead of extra-repo. I would connect to arduino not only DHT things, but other as well.

soundanalogous:
There are 2 reasons I'd like it to be in a separate repo:
1. This is a device as opposed to a capability (such as PWM, I2C, SPI, etc).
2. It wraps a 3rd party dht library. Ideally OneWire and Stepper would be removed from the ConfigurableFirmata core since they also wrap 3rd party libraries but they have to remain for backwards compatibility.

Also look into including the Adafrult dht library instead since it covers a wider variety of dht sensors: https://github.com/adafruit/DHT-sensor-library and it also uses shorter fixed delays than the version that was previously included in the DHT firmata example.

soundanalogous:
The adafruit DHT library is also available via the Arduino Library manager.

PR #247, opened 2015-12-08
Make VERSION_BLINK_PIN optional

cmaglie:
This may be useful for boards without a built-in led or if you simply want to disable this feature.

Also removed a redundant `pinMode` (the pin is already set as output inside `strobeBlinkPin`).

### DIFF
--- a/Firmata.cpp
+++ b/Firmata.cpp
@@ -94,7 +94,6 @@ void FirmataClass::printVersion(void)
 void FirmataClass::blinkVersion(void)
 {
   // flash the pin with the protocol version
-  pinMode(VERSION_BLINK_PIN, OUTPUT);
   strobeBlinkPin(FIRMATA_MAJOR_VERSION, 40, 210);
   delay(250);
   strobeBlinkPin(FIRMATA_MINOR_VERSION, 40, 210);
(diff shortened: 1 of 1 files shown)


soundanalogous:
I'd rather have `blinkVersion()` do nothing in this case to avoid the calls to `delay`. Add the compiler conditional inside `blinkVersion()` instead so it essentially does nothing if not LED pin is specified. This should be safe since `strobeBlinkPin()` is private and is only called by `blinkVersion()`.

soundanalogous:
@cmaglie I've made an [alternative version](https://github.com/firmata/arduino/pull/256) that avoids the delays in blinkVersion. However I'm not sure how the feature can be disabled as you requested, other than the user modifying Boards.h to undefine `VERSION_BLINK_PIN` for a particular board. Let me know if you have any ideas. I tried adding the following [here](https://github.com/firmata/arduino/blob/optional-blink/Firmata.cpp#L96)
```c++
#if defined(VERSION_BLINK_PIN) && !defined(FIRMATA_SKIP_BLINK_VERSION)
```
and then defining `FIRMATA_SKIP_BLINK_VERSION` in a sketch before including `Firmata.h` but that doesn't work. I assume `Firmata.h` is compiled before any defines above it in a sketch are evaluated by the preprocessor. Perhaps there is something simple I am overlooking.

PR #256, opened 2015-12-21
make VERSION_BLINK_PIN optional

soundanalogous:
This is an alternative approach to: https://github.com/firmata/arduino/pull/247/files. However neither approach allows the user to override `VERSION_BLINK_PIN` if it is specified already in Boards.h for the board they are compiling against. This is because `VERSION_BLINK_PIN` is defined per board (that have an onboard LED) in Boards.h and Firmata.cpp includes Boards.h, so there is no way to `undef` `VERSION_BLINK_PIN` from the user's sketch.

### DIFF
--- a/Firmata.cpp
+++ b/Firmata.cpp
@@ -93,12 +93,14 @@ void FirmataClass::printVersion(void)
 
 void FirmataClass::blinkVersion(void)
 {
+#if defined(VERSION_BLINK_PIN)
   // flash the pin with the protocol version
   pinMode(VERSION_BLINK_PIN, OUTPUT);
-  strobeBlinkPin(FIRMATA_MAJOR_VERSION, 40, 210);
+  strobeBlinkPin(VERSION_BLINK_PIN, FIRMATA_MAJOR_VERSION, 40, 210);
   delay(250);
-  strobeBlinkPin(FIRMATA_MINOR_VERSION, 40, 210);
+  strobeBlinkPin(VERSION_BLINK_PIN, FIRMATA_MINOR_VERSION, 40, 210);
   delay(125);
+#endif
 }
 
 void FirmataClass::printFirmwareVersion(void)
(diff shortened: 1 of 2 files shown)


cmaglie:
:+1:  LGTM

> However I'm not sure how the feature can be disabled as you requested, other than the user modifying Boards.h to undefine VERSION_BLINK_PIN for a particular board.

Yes, the user should change an existing board or define his own. BTW this is mainly intended for boards without LEDs.

Thanks for taking the time to look at this one!

Firmata.cpp
```diff
@@ -93,12 +93,14 @@ void FirmataClass::printVersion(void)
...
 void FirmataClass::blinkVersion(void)
 {
+#if defined(VERSION_BLINK_PIN)
   // flash the pin with the protocol version
   pinMode(VERSION_BLINK_PIN, OUTPUT);
-  strobeBlinkPin(FIRMATA_MAJOR_VERSION, 40, 210);
+  strobeBlinkPin(VERSION_BLINK_PIN, FIRMATA_MAJOR_VERSION, 40, 210);
   delay(250);
-  strobeBlinkPin(FIRMATA_MINOR_VERSION, 40, 210);
+  strobeBlinkPin(VERSION_BLINK_PIN, FIRMATA_MINOR_VERSION, 40, 210);
```
rwaldron: This is a nitpick, but should this also be `FIRMATA_` prefixed?

soundanalogous: Ideally but it has been defined in Boards.h for years. I'm not sure if anyone is using it in their sketches - probably not because it is only used in Firmata.cpp, but nothing prevents anyone from using it in a sketch as well. This is a issue I've been having lately. Each new Arduino board has included a duplicate define somewhere in it's code base that has broken something in Firmata so prefixing is necessary... just trying to figure out a way to do it without breaking existing sketches.

PR #263, opened 2016-01-08
Include Extra Functionality

DomAmato:
Jeff, this just makes it so that features like Encoders, Steppers, and Onewire can be easily added in. I know that configurableFirmata generally covers these features but with these files adding in funcitonality is as easy as including or not including the header. See this for reference:
https://github.com/brinnLabs/Robust-Firmata/blob/master/RobustFirmata/RobustFirmata.ino

I modeled it after the Serial Utility addition

### DIFF
--- a/utility/Encoder.cpp
+++ b/utility/Encoder.cpp
@@ -1,0 +1,10 @@
+
+#include "Encoder.h"
+
+// Yes, all the code is in the header file, to provide the user
+// configure options with #define (before they include it), and
+// to facilitate some crafty optimizations!
+
+Encoder_internal_state_t * Encoder::interruptArgs[];
+
+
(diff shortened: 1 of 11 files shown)


DomAmato:
I suppose people could just install the respective libraries and use the same method so maybe this isn't ideal though it prevents 3rd party library updates from breaking things when they are updated. Maybe this is more food for thought... Also i realize I added in the stepper changes I made for my own purposes but that are still open issues on the protocol repo

soundanalogous:
All of these libraries should be installed separately. Also some boards are including libraries like OneWire in their own distribution (Arduino 101 for example) and this will likely lead to conflicts.

soundanalogous:
I do plan to add I2CUtility to move the I2C related defines out of StandardFirmata*. I may also add transport classes to sort out the various transports (other than Serial) such as Ethernet, WiFi and BLE.

soundanalogous:
Also I'm going to change the OneWire Firmata implementation slightly so that it no longer relies on the modified version of OneWire. It should work with the official version that users can install to their sketchbook libraries folder.

DomAmato:
Cool, I opened an issue yesterday with onewire as well since the arduino library is out of sync with the repo and Paul said to open an issue elsewhere.

soundanalogous:
I just checked OneWire in the Library manager. You need to click on the OneWire item in the list, then click on Update.

PR #268, opened 2016-01-11
Serial feature class

soundanalogous:
#### Broke out Firmata serial feature code to new class
- Added SerialFirmata class and FirmataFeature interface
- Update all StandardFirmata variants to use pinMode and pinState
  getters and setters that were added to Firmata.cpp in the last
  commit.

In this way, feature classes from ConfigurableFirmata can be used with Firmata. SerialFirmata was copied over from ConfigurableFirmata with minimal changes. Simplifies maintenance of both libraries.

Greatly simplifies maintenance of StandardFirmata variants.

#### Refactored pin state and pin config handling
- Moved pin config handling to Firmata.cpp
- Moved pin state handling to Firmata.cpp
- Added documentation for all methods in Firmata.cpp

The pin config and pin state changes enable Firmata features to be extracted from StandardFirmata into their own classes.


soundanalogous:
I plan to break out the I2C and Servo features to their own classes as well. Will be in a separate PR since Serial was priority.

soundanalogous:
change `setPinMode` and `getPinMode` back to `setPinConfig` and `getPinConfig` to avoid confusion. Should add `setPinConfig` and `getPinConfig` to ConfigurableFirmata as well and deprecate the `setPinMode` and `getPinMode`.

Issue #260, opened 2016-01-02
Changing baud rate in softwareSerial


dtex:
Moved from https://github.com/jgautier/firmata/issues/104

### My Scenario

Using an Arduino Uno and softwareSerial I'm trying to connect to a device that runs at 115200 baud by default. I know the Uno & softwareSerial cannot handle that baud rate reliably so the first thing I do is send a short command to change the device to 9600 baud, then close my connection and reopen at 9600. 

### Sample Arduino code

This sample works fine:

````C
#include <SoftwareSerial.h>
SoftwareSerial mySerial(11, 10); // RX, TX

void setup() {
  // Open serial communications and wait for port to open:
  Serial.begin(57600);
  
  while (!Serial) {
    ; // wait for serial port to connect. Needed for native USB port only
  }

mySerial.begin(115200);
  
  // Send 32 byte command to change message type and baud rate
  byte changeOpMode[] = {0xA0, 0xA2, 0x00, 0x18, 0x81, 0x02, 0x01, 0x01, 0x00, 0x01, 0x01, 0x01, 0x05, 0x01, 0x01, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x25, 0x80, 0x01, 0x3A, 0xB0, 0xB3 };
  mySerial.write(changeOpMode, sizeof(changeOpMode));
  mySerial.end();
  
  // Reopen connection at new baud rate
  mySerial.begin(9600);
}

void loop() { // run over and over
  if (mySerial.available()) {
    Serial.write(mySerial.read());
  }
  if (Serial.available()) {
    mySerial.write(Serial.read());
  }
}
````

> @soundanalogous wrote:
> The root of this issue is on the Arduino (c/c++) side, not the JavaScript side. I can see now based on your Arduino example above what is different from my implementation. I'm deallocating the pointer to the SoftwareSerial object when calling end(), then if the user tries to config again, a new SoftwareSerial instance is assigned to the pointer that must be breaking something (and could also lead to a heap and stack collision if done multiple times). Whereas in your example above you simply call end, then begin with a new baud rate.

> I want to keep the ability to deallocate memory in case a serial port is no longer needed after a certain point in an application. What I could do is add a new message that enables the user to update the baud rate. This would essentially call end then begin with the new baud as in your example. The close would deallocate.

soundanalogous:
[...]

PR #270, opened 2016-01-17
enable baud rate for serial feature to be updated

soundanalogous:
This is useful when you need to change the baud rate of a serial peripheral.

### DIFF
--- a/utility/SerialFirmata.cpp
+++ b/utility/SerialFirmata.cpp
@@ -14,7 +14,7 @@
 
   - handlePinMode calls Firmata::setPinMode
 
-  Last updated by Jeff Hoefs: January 10th, 2016
+  Last updated by Jeff Hoefs: January 16th, 2016
 */
 
 #include "SerialFirmata.h"
@@ -199,6 +199,23 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
         }
         break; // SERIAL_LISTEN
 #endif
+      case SERIAL_UPDATE_BAUD:
+        serialPort = getPortFromId(portId);
+        if (serialPort == NULL) {
+          break;
+        }
+        long newBaud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
+        if (portId < 8) {
+          ((HardwareSerial*)serialPort)->end();
+          delay(100); // delay required for HW serial only
+          ((HardwareSerial*)serialPort)->begin(newBaud);
+        } else {
+#if defined(SoftwareSerial_h)
+          ((SoftwareSerial*)serialPort)->end();
+          ((SoftwareSerial*)serialPort)->begin(newBaud);
+#endif
+        }
+        break; // SERIAL_UPDATE_BAUD
     } // end switch
     return true;
   }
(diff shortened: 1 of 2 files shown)


utility/SerialFirmata.cpp
```diff
@@ -199,6 +199,23 @@ boolean SerialFirmata::handleSysex(byte command, byte argc, byte *argv)
         }
         break; // SERIAL_LISTEN
 #endif
+      case SERIAL_UPDATE_BAUD:
+        serialPort = getPortFromId(portId);
+        if (serialPort == NULL) {
+          break;
+        }
+        long newBaud = (long)argv[1] | ((long)argv[2] << 7) | ((long)argv[3] << 14);
+        if (portId < 8) {
```
soundanalogous: consider calling `flush()` before calling `end()`

dtex:
Jeff,

I'm sorry to do this to you but how do I update the Arduino IDE to use the source from the repo? You once sent me a zip (when testing something else) that I dropped into the app package (OS X) and I'm trying to do that with the repo, but I must be putting it in the wrong path. I even deleted the existing firmata folder at ```content/java/libraries/``` but firmata still shows up in examples. I'm using 1.6.5

soundanalogous:
There is no need to copy Firmata into the Arduino IDE bundle as long as you are using Arduino IDE 1.6.4 or higher.

Simply copy (or clone) Firmata (renaming to `Firmata` when you clone) into you sketchbook libraries directory.

dtex:
I tried many times and could not get a successful switch to 9600 (based on the data I'm getting back).

Here's my test code. As best I can tell it should behave exactly like [this example](https://github.com/firmata/arduino/issues/260#issuecomment-168435942). I added the timeouts just to be conservative:

````js
var firmata = require("firmata");
var board = new firmata("/dev/cu.usbmodem14231");

board.on("ready", function() {
  
  board.serialConfig({portId: board.SERIAL_PORT_IDs.HW_SERIAL1, baud: 115200});

  setTimeout( function() {
    board.serialRead(board.SERIAL_PORT_IDs.HW_SERIAL1, function(data) {
      console.log("read", data);
    });
  }, 1000);

  setTimeout( function() {
    board.serialWrite(board.SERIAL_PORT_IDs.HW_SERIAL1, [0xA0, 0xA2, 0x00, 0x18, 0x81,
      0x02, 0x01, 0x01, 0x00, 0x01, 0x01, 0x01, 0x05, 0x01, 0x01, 0x01,
      0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x00, 0x01, 0x25,
      0x80, 0x01, 0x3A, 0xB0, 0xB3]);
  }, 2000);

  setTimeout( function() {
    board.serialBaud(board.SERIAL_PORT_IDs.HW_SERIAL1, 9600);
  }, 3000);

});
````

soundanalogous:
Did you check out the [update-baud branch](https://github.com/firmata/arduino/tree/update-baud) before uploading StandardFirmataPlus to your board?

soundanalogous:
I know what the problem is. I made a really stupid mistake in the protocol. I can't use `0x80` (for `SERIAL_UPDATE_BAUD`) as a command byte because it's > 127 decimal. Luckily I left `0x00` open so I can use that to signal an extended command set. I should really start allocating 2 bytes for commands from now on. I'll have to go back to the protocol stage to fix this.

soundanalogous:
Here's one solution: https://github.com/firmata/protocol/pull/46#issuecomment-172748196.

soundanalogous:
I'm going to release a SerialFirmata 2.0 at some point that fixes the larger scalability issue here and will then revisit the ability to update the serial baud rate.

Issue #193, opened 2015-03-17
provide an option to free up TX and RX pins when using Ethernet transport


soundanalogous:
Now that Ethernet is supported, provide an option to free up the RX and TX pins. The could be used with [Firmata Serial](https://github.com/firmata/protocol/blob/serial/serial-proposal.md) or as normal digital I/O pins.

soundanalogous:
stretch goal for 2.6 release

soundanalogous:
This is complicated by the need to keep a UART available as a convenience for debugging when using Ethernet or Wi-Fi transports (especially for Wi-Fi).

PR #278, opened 2016-03-13
pin functions based on board depended defaults

jnsbyr:
to support different ESP8266 boad layouts, use board depended constants
from "...\esp8266\hardware\esp8266\2.1.0\variants\...\pins_arduino.h"

DOUT, DIN, I2C and SERIAL0 have been tested and are working with a HUZZA board

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -685,7 +685,7 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define IS_PIN_INTERRUPT(p)     (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 15))
 #define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)
 #define PIN_TO_DIGITAL(p)       (p)
-#define PIN_TO_ANALOG(p)        (p)
+#define PIN_TO_ANALOG(p)        ((p) - 17)
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p)
 
--- a/Boards.h
+++ b/Boards.h
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         ((p) - 2)
 
-// ESP8266 generic
+// ESP8266
+// hardware: TX0=1, TX1=2, RX0=3, MISO=12, MOSI=13, SCLK=14, CS=15, A0=17  
+// board depended defaults: LED_BUILTIN, SDA, SCL
+// note: boot mode GPIOs 0, 2 and 15 can be used as outputs (GPIOs 6-11 are in use for flash IO)
 #elif defined(ESP8266)
-#define TOTAL_ANALOG_PINS       0
-#define TOTAL_PINS              17
-#define VERSION_BLINK_PIN       4
-// #define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 1 || (p) == 2 || (p) == 3 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16) //for wifi dont protect serial pins because these things only have 2 pins otherwise
-#define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 2 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16)
-#define IS_PIN_ANALOG(p)        (false)
-#define IS_PIN_PWM(p)           (false)
-#define IS_PIN_SERVO(p)         ((p) >= 0 && (p) < MAX_SERVOS)
-#define IS_PIN_I2C(p)           (false)
-#define IS_PIN_SPI(p)           (false)
+#define TOTAL_ANALOG_PINS       1
+#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define PIN_SERIAL_RX           3
+#define PIN_SERIAL_TX           1
+#define PIN_SERIAL1_TX          2
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
+#define IS_PIN_ANALOG(p)        ((p) == A0)
+#define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
+#define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
+#define IS_PIN_SPI(p)           ((p) >= 12 && (p) <= 15)
+#define IS_PIN_INTERRUPT(p)     (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 15))
+#define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)
 #define PIN_TO_DIGITAL(p)       (p)
-#define PIN_TO_ANALOG(p)        ((p) - 17)
+#define PIN_TO_ANALOG(p)        (p)
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
-#define PIN_TO_SERVO(p)         p
+#define PIN_TO_SERVO(p)         (p)
 
  
 // anything else
--- a/Boards.h
+++ b/Boards.h
@@ -666,24 +666,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         ((p) - 2)
 
 // ESP8266
-// hardware: TX0=1, TX1=2, RX0=3, MISO=12, MOSI=13, SCLK=14, CS=15, A0=17  
-// board depended defaults: LED_BUILTIN, SDA, SCL
-// note: boot mode GPIOs 0, 2 and 15 can be used as outputs (GPIOs 6-11 are in use for flash IO)
+// note: boot mode GPIOs 0, 2 and 15 can be used as outputs, GPIOs 6-11 are in use for flash IO
 #elif defined(ESP8266)
-#define TOTAL_ANALOG_PINS       1
-#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0 + NUM_ANALOG_INPUTS)
+#define VERSION_BLINK_PIN       LED_BUILTIN
 #define PIN_SERIAL_RX           3
 #define PIN_SERIAL_TX           1
-#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
-#define IS_PIN_ANALOG(p)        ((p) == A0)
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) < (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0)))
+#define IS_PIN_ANALOG(p)        ((p) >= A0 && (p) < A0 + NUM_ANALOG_INPUTS)
 #define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)
 #define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
 #define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
-#define IS_PIN_SPI(p)           ((p) >= 12 && (p) <= 15)
-#define IS_PIN_INTERRUPT(p)     (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 15))
-#define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)
+#define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
+#define IS_PIN_INTERRUPT(p)     digitalPinToInterrupt(p)
+#define IS_PIN_SERIAL(p)        ((p) == PIN_SERIAL_RX || (p) == PIN_SERIAL_TX)
 #define PIN_TO_DIGITAL(p)       (p)
-#define PIN_TO_ANALOG(p)        ((p) - 17)
+#define PIN_TO_ANALOG(p)        ((p) - A0)
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p)
 
--- a/Boards.h
+++ b/Boards.h
@@ -672,10 +672,8 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #elif defined(ESP8266)
 #define TOTAL_ANALOG_PINS       1
 #define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
-#define VERSION_BLINK_PIN       LED_BUILTIN
 #define PIN_SERIAL_RX           3
 #define PIN_SERIAL_TX           1
-#define PIN_SERIAL1_TX          2
 #define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
 #define IS_PIN_ANALOG(p)        ((p) == A0)
 #define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)
--- a/Boards.h
+++ b/Boards.h
@@ -669,13 +669,12 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 // note: boot mode GPIOs 0, 2 and 15 can be used as outputs, GPIOs 6-11 are in use for flash IO
 #elif defined(ESP8266)
 #define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
-#define TOTAL_PINS              (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0 + NUM_ANALOG_INPUTS)
-#define VERSION_BLINK_PIN       LED_BUILTIN
+#define TOTAL_PINS              A0 + NUM_ANALOG_INPUTS
 #define PIN_SERIAL_RX           3
 #define PIN_SERIAL_TX           1
-#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) < (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0)))
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) < A0))
 #define IS_PIN_ANALOG(p)        ((p) >= A0 && (p) < A0 + NUM_ANALOG_INPUTS)
-#define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
 #define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
 #define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
 #define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
--- a/Boards.h
+++ b/Boards.h
@@ -678,7 +678,7 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
 #define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
 #define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
-#define IS_PIN_INTERRUPT(p)     digitalPinToInterrupt(p)
+#define IS_PIN_INTERRUPT(p)     (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)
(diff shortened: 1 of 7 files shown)


Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
-// #define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 1 || (p) == 2 || (p) == 3 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16) //for wifi dont protect serial pins because these things only have 2 pins otherwise
-#define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 2 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16)
-#define IS_PIN_ANALOG(p)        (false)
-#define IS_PIN_PWM(p)           (false)
-#define IS_PIN_SERVO(p)         ((p) >= 0 && (p) < MAX_SERVOS)
-#define IS_PIN_I2C(p)           (false)
-#define IS_PIN_SPI(p)           (false)
+#define TOTAL_ANALOG_PINS       1
+#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define VERSION_BLINK_PIN       LED_BUILTIN
```
soundanalogous: Remove `VERSION_BLINK_PIN`. The version blink sequence takes too long and excluding this define will exclude the version blink sequence - saving up to 2.5 seconds or more of startup time (since each Firmata version increment adds additional time to the sequence).

Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define PIN_SERIAL_RX           3
+#define PIN_SERIAL_TX           1
+#define PIN_SERIAL1_TX          2
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
+#define IS_PIN_ANALOG(p)        ((p) == A0)
+#define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
+#define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
```
soundanalogous: Is SDA and SCL defined for all ESP variants or only specific ones?

Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         ((p) - 2)
 
-// ESP8266 generic
+// ESP8266
```
soundanalogous: There are several different popular boards using this chip. We should include a comment that states which specific boards have been tested to work with this set of pin definitions.

Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
-#define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 2 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16)
-#define IS_PIN_ANALOG(p)        (false)
-#define IS_PIN_PWM(p)           (false)
-#define IS_PIN_SERVO(p)         ((p) >= 0 && (p) < MAX_SERVOS)
-#define IS_PIN_I2C(p)           (false)
-#define IS_PIN_SPI(p)           (false)
+#define TOTAL_ANALOG_PINS       1
+#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define PIN_SERIAL_RX           3
```
soundanalogous: Just to verify - this maps to `Serial` (as opposed to `Serial1`) right?

Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define PIN_SERIAL_TX           1
+#define PIN_SERIAL1_TX          2
```
soundanalogous: I don't think Firmata Serial supports just a TX pin without the RX pair.

Boards.h
jnsbyr: will be removed with next commit

Boards.h
jnsbyr: I thought so and that is way I choose to set ```#define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)```. It's more to be complete about the theoretical possibilities but maybe it is better to remove the PIN_SERIAL1_TX.

Boards.h
jnsbyr: Pin 3+1 are Serial (or port 0) and can be used for SerialFirmata.

Boards.h
jnsbyr: I don't know all these boards myself, but I checked all the "variants...\pins_arduino.h" header files. They all have the pins defined, mostly on pins 4+5, but not all. Maybe the SDA/SCL pins are not always physically available to the user. From the project documentation I get the impression that SPI is always available. The pins in the header files are defaults but can be remapped to almost any other pins with Wire.begin(...) similar to SoftwareSerial.

jnsbyr:
Note that with current release V2.1 of esp8266/Arduino you need 
```C
#if !defined(ARDUINO_ARCH_ESP8266)
          Firmata.sendAnalog(analogPin, analogRead(analogPin));
#else
          Firmata.sendAnalog(analogPin, analogRead(A0));
#endif        
```
to make the analog input work. I opened an issue (https://github.com/esp8266/Arduino/issues/1766) to make the define unnecessary.

Boards.h
```diff
@@ -665,22 +665,29 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
+#define IS_PIN_ANALOG(p)        ((p) == A0)
```
rwaldron: What about the wifio?

```
static const uint8_t A0 = 14;
static const uint8_t A1 = 15;
static const uint8_t A2 = 16;
static const uint8_t A3 = 17;
static const uint8_t A4 = 18;
static const uint8_t A5 = 19;
static const uint8_t A6 = 20;
static const uint8_t A7 = 21;
```

?

Boards.h
rwaldron: > mostly on pins 4+5, but not all

Confirmed. 

Sparkfun Thing: 
```
static const uint8_t SDA = 2;
static const uint8_t SCL = 14;
```

(The only exception)

Boards.h
jnsbyr: Please tell me where these constants are needed for. A0 = 17 is already in use in the esp8266/Arduino project for the one and only analog input pin.

Boards.h
```diff
@@ -665,22 +665,27 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
-// #define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 1 || (p) == 2 || (p) == 3 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16) //for wifi dont protect serial pins because these things only have 2 pins otherwise
-#define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 2 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16)
-#define IS_PIN_ANALOG(p)        (false)
-#define IS_PIN_PWM(p)           (false)
-#define IS_PIN_SERVO(p)         ((p) >= 0 && (p) < MAX_SERVOS)
-#define IS_PIN_I2C(p)           (false)
-#define IS_PIN_SPI(p)           (false)
+#define TOTAL_ANALOG_PINS       1
+#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define PIN_SERIAL_RX           3
```
soundanalogous: `PIN_SERIAL_RX` and `PIN_SERIAL_TX` are no-ops: https://github.com/firmata/arduino/blob/master/utility/SerialFirmata.cpp#L237-L240. No harm in including them here but the use of `Serial` in SerialFirmata is not yet supported due to possible conflicts with boards that use `Serial` as the primary communication with Firmata.

soundanalogous: per this issue: https://github.com/firmata/arduino/issues/193

Boards.h
jnsbyr: OK, found it myself. For this board probably a separate definition is needed in Board.h and that may not be enough to make it work.

Boards.h
jnsbyr: Forgot about this, cause I'm mixing Firmata, ConfigurableFirmata and my own mods of Firmata a lot. So this is for future use.

soundanalogous:
As much as I hate to add yet another StandardFirmata variant, we may need to add StandardFirmataESP and include espConfig.h (much in the same way I've included a config for Ethernet, WiFi and BLE). That way the user can uncomment a define that will configure the sketch properly for their particular board.

soundanalogous:
Or it could all be handled in wifiConfig.h (in StandardFirmataWiFi) - provide different configurations and the user selects one by uncommenting an appropriate define.

soundanalogous:
I see now the wifiConfig.h route is what [Jacob had started](https://github.com/firmata/arduino/commit/9586defdb5fc711f0343add8978b7f1feccaf6a4#diff-08f7b758e2fa8d05d6c85b959d53e36dR64) so lets continue down that route. Also need to figure out a better way to manage [this](https://github.com/firmata/arduino/commit/9586defdb5fc711f0343add8978b7f1feccaf6a4#diff-08f7b758e2fa8d05d6c85b959d53e36dR143) as it's starting to get out of hand.

soundanalogous:
One thing that would make this much easier is if [each variant in esp8266/Arduino](https://github.com/esp8266/Arduino/tree/master/variants) had a unique define - then we could sort these out easily in Boards.h.

jnsbyr:
For my tests I didn't need to go so far as the changes you refer to. I based my tests on StandardFirmataEthernetPlus, changed Ethernet into WiFi, added a few lines to setup WiFi and added a few defines to exclude the Arduino board special pin configurations.

Having a unique identifier for the different boards would help. Maybe the project is open to such an enhancement.

soundanalogous:
One way to have a different unique identifier is to have a unique sketch name for each board and inspect that name view the firmware query (which returns the filename)

soundanalogous:
I looked over the `pins_arduino.h` files for all of the variants in esp8266/Arduino. The following should cover everything except the wifio board. The serial pin mapping may not be correct for all boards. Also some digital pins may need to be ignored (and can be in wifiConfig.h if necessary):

```
#elif defined(ESP8266)
#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
#define TOTAL_PINS              A0 + NUM_ANALOG_INPUTS
#define PIN_SERIAL_RX           3
#define PIN_SERIAL_TX           1
#define IS_PIN_DIGITAL(p)       ((p) >= 0 && (p) < A0)
#define IS_PIN_ANALOG(p)        ((p) >= A0 && (p) < A0 + NUM_ANALOG_PINS)
#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
#define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
#define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
#define IS_PIN_INTERRUPT(p)     digitalPinToInterrupt(p)
#define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)
#define PIN_TO_DIGITAL(p)       (p)
#define PIN_TO_ANALOG(p)        ((p) - A0)
#define PIN_TO_SERVO(p)         (p)
```

jnsbyr:
Looks good to me, but I would explicitly exclude the pins 6 to 11. They are used for the flash IO on most of the boards and accessing them by accident will probably crash the application. Using the higher number of the total pins constant from pins_arduino.h for an ifdef one could try to make a special config for wifio.

Boards.h
rwaldron: :+1:

soundanalogous:
Untested, but this may cover all boards:

```
#elif defined(ESP8266)
#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
#define TOTAL_PINS              (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0 + NUM_ANALOG_INPUTS)
#define PIN_SERIAL_RX           3
#define PIN_SERIAL_TX           1
#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) < (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0)))
#define IS_PIN_ANALOG(p)        ((p) >= A0 && (p) < A0 + NUM_ANALOG_PINS)
#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS)
#define IS_PIN_I2C(p)           ((p) == SDA || (p) == SCL)
#define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
#define IS_PIN_INTERRUPT(p)     digitalPinToInterrupt(p)
#define IS_PIN_SERIAL(p)        ((p) == 1 || (p) == 3)
#define PIN_TO_DIGITAL(p)       (p)
#define PIN_TO_ANALOG(p)        ((p) - A0)
#define PIN_TO_SERVO(p)         (p)
```

jnsbyr:
Don't know why but I had trouble with the line ```#define IS_PIN_PWM(p)           digitalPinHasPWM(p)```. The capability query returned all pins and not only the digital pins so I changed it into ```#define IS_PIN_PWM(p)           IS_PIN_DIGITAL(p)```. Maybe the xtensa compiler works a little bit different when resolving macros.

jnsbyr:
Because of https://github.com/esp8266/Arduino/issues/1766 ```#define PIN_TO_ANALOG(p)        ((p) - A0)``` will not help with wifio A1 to A7. Either the esp8266/Arduino project also remaps these pins for ```analogRead()``` or one adds a special reverse mapping to Firmata for calling ```analogRead()```.

soundanalogous:
I don't see how that would be an issue since wifio A0 is mapped to pin 14 as it should be for that board: https://github.com/esp8266/Arduino/blob/master/variants/wifio/pins_arduino.h#L41. 

`PIN_TO_ANALOG` should work correctly since 14 - 14 = 0, 15 - 14 = 1, etc.

jnsbyr:
I did not say that it's not possible but that the esp project V2.1 assumes that analogRead is called with the absolute pin number. Even A0 is currently only working by calling analogRead(17) but PIN_TO_ANALOG would call analogRead(0) instead.

soundanalogous:
The broken pin mapping in esp8266 is definitely a blocker. Hopefully they resolve that soon.

igrr:
Guys, you may ignore wifio variant — it's a board which was never released. We keep the code in repository because it shows how to implement "virtual" IO with an AVR slave. Noone is actually using it.

I'll fix analogRead(0) asap.

rwaldron:
@igrr thanks for the details :)

jnsbyr:
* ```#define IS_PIN_PWM(p)           digitalPinHasPWM(p)``` results in all pins including the analog pin claiming to have PWM capability. The same probably applies to ```#define IS_PIN_INTERRUPT(p)     digitalPinToInterrupt(p)``` but that macro is not used in Firmata afaik
* Firmata 2.5 does not compile without ```VERSION_BLINK_PIN``` defined
* analog pin A0 works with current git version of "cores/esp8266/core_esp8266_wiring_analog.c" on Huzza board

soundanalogous:
> #define IS_PIN_PWM(p) digitalPinHasPWM(p) results in all pins including the analog pin claiming to have PWM capability. The same probably applies to #define IS_PIN_INTERRUPT(p) digitalPinToInterrupt(p) but that macro is not used in Firmata afaik

lets try to get the `digitalPinHasPWM(p)` and `digitalPinToInterrupt(p)` macros fixed in esp8266/Arduino

> Firmata 2.5 does not compile without VERSION_BLINK_PIN defined

It should compile in 2.5.1 or higher. I removed the dependency on `VERSION_BLINK_PIN` in [v2.5.1](https://github.com/firmata/arduino/blob/master/extras/revisions.txt#L19).

jnsbyr:
> lets try to get the digitalPinHasPWM(p) and digitalPinToInterrupt(p) macros fixed in esp8266/Arduino

Agreed, ```#define digitalPinHasPWM(p)         (((p) < NUM_DIGITAL_PINS)?p:NOT_A_PIN)``` from 
file pins_arduino.h result in the pin list "1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17" for the Huzza and that is not right, cause pin numbering on the ESP starts with 0 and 17 is the analog pin. 

Upgraded to Firmata 2.5.2 and now I can do without ```VERSION_BLINK_PIN```.

Boards.h
```diff
@@ -666,24 +666,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
 // ESP8266
-// hardware: TX0=1, TX1=2, RX0=3, MISO=12, MOSI=13, SCLK=14, CS=15, A0=17  
-// board depended defaults: LED_BUILTIN, SDA, SCL
-// note: boot mode GPIOs 0, 2 and 15 can be used as outputs (GPIOs 6-11 are in use for flash IO)
+// note: boot mode GPIOs 0, 2 and 15 can be used as outputs, GPIOs 6-11 are in use for flash IO
 #elif defined(ESP8266)
-#define TOTAL_ANALOG_PINS       1
-#define TOTAL_PINS              18 // 11 digital + 1 analog + 6 inaccessible
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0 + NUM_ANALOG_INPUTS)
```
soundanalogous: the conditional isn't needed anymore, should just be `A0 + NUM_ANALOG_INPUTS`.

Boards.h
```diff
@@ -666,24 +666,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define VERSION_BLINK_PIN       LED_BUILTIN
 #define PIN_SERIAL_RX           3
 #define PIN_SERIAL_TX           1
-#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) <= 16))
-#define IS_PIN_ANALOG(p)        ((p) == A0)
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 5) || ((p) >= 12 && (p) < (NUM_DIGITAL_PINS > A0 + NUM_ANALOG_INPUTS ? NUM_DIGITAL_PINS : A0)))
```
soundanalogous: should be able to change back to: `#define IS_PIN_DIGITAL(p)       ((p) >= 0 && (p) < A0)`

jnsbyr:
Added a request to fix the macros ```digitalPinHasPWM``` and ```digitalPinToInterrupt``` to issue 1766 of esp8266/Aduino (see https://github.com/esp8266/Arduino/issues/1766#issuecomment-198565416).

jnsbyr:
The ```digitalPinHasPWM``` macro has been fixed in the current Github version and is staged for release with version 2.2.0, see https://github.com/esp8266/Arduino/issues/1831#issuecomment-204451341. This should complete the changes to Boards.h for the ESP8266 variants.

jnsbyr:
Just found something else not considered so far: the ESP8266 project default PWM resolution is 10 bits, not 8, but Firmata always claims 8 bit PWM resolution, even with AVR hardware that can do better (Due, Zero), relying on API defaults. Using ```analogWriteResolution``` or ```analogWriteRange``` the resolution may be modified. The behaviour is probably platform depended.

soundanalogous:
I drafted a proposal to add analog configuration a couple of years ago: https://github.com/firmata/protocol/pull/8

I also opened a PR for this long ago but ran into MCU architecture issues that I was not able to resolve: https://github.com/firmata/arduino/pull/174

Open to picking this back up if I can find the time.

jnsbyr:
#174 adds interesting configuration options but they go much further than necessary for basic operations. A capability query should not return the wrong properties. Even if a platform does not support changing PWM resolution, it should report the build-in resolution and on more powerful platforms the default resolution.

A simple but ugly way to do it is 
```C
        if (IS_PIN_PWM(pin)) {
          Firmata.write(PIN_MODE_PWM);
#if defined(ARDUINO_ARCH_ESP8266)
          Firmata.write(10); // 10 = 10-bit resolution, see Arduino.h PWMRANGE
#else          
          Firmata.write(8);  // 8 = 8-bit resolution
#endif          
        }
```
but a new property like ```#define DEFAULT_PWM_RESOLUTION 8``` in Boards.h would provide the basic requirements without a protocol enhancement:

```C
        if (IS_PIN_PWM(pin)) {
          Firmata.write(PIN_MODE_PWM);
          Firmata.write(DEFAULT_PWM_RESOLUTION);  // resolution in bits
        }
```

soundanalogous:
I think the second approach is a good short-term solution. Some Arduino variants already report the default pwm resolution ([for example](https://github.com/arduino/Arduino/blob/master/hardware/arduino/sam/variants/arduino_due_x/variant.h#L223)), but it's probably irregular enough to justify adding it to Boards.h. Can you add `DEFAULT_PWM_RESOLUTION`? Then I think this PR will be ready to merge.

Boards.h
```diff
@@ -665,22 +665,25 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
-// #define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 1 || (p) == 2 || (p) == 3 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16) //for wifi dont protect serial pins because these things only have 2 pins otherwise
-#define IS_PIN_DIGITAL(p)       ((p) == 0 || (p) == 2 || (p) == 4 || (p) == 5 || (p) == 12 || (p) == 13 || (p) == 14 || (p) == 15 || (p) == 16)
-#define IS_PIN_ANALOG(p)        (false)
-#define IS_PIN_PWM(p)           (false)
-#define IS_PIN_SERVO(p)         ((p) >= 0 && (p) < MAX_SERVOS)
-#define IS_PIN_I2C(p)           (false)
-#define IS_PIN_SPI(p)           (false)
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              A0 + NUM_ANALOG_INPUTS
+#define PIN_SERIAL_RX           3
```
soundanalogous: Do pins 3 and 1 appear to be common for RX and TX across all ESP8266 board variants?

jnsbyr: Serial has some special properties with the ESP8266. I don't think they can be completely described in a Boards.h approach (e.g. similar to the PWM resolution you can change the serial pin assignment and there is a Serial1 but you typically can't use the RX pin). The details are described [here](https://github.com/esp8266/Arduino/blob/master/doc/reference.md#serial).

All boards seem to have Serial in common and the default mapping for Serial is always pins 1 and 3.

PR #285, opened 2016-04-09
WiFiStream variant integration

jnsbyr:
```STATIC_IP_ADDRESS``` is currently not commented out for testing purposes in _wifiConfig.h_

### DIFF
--- a/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
+++ b/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
@@ -125,6 +125,16 @@ SerialFirmata serialFeature;
 #ifdef STATIC_IP_ADDRESS
 IPAddress local_ip(STATIC_IP_ADDRESS);
 #endif
+#ifdef SUBNET_MASK
+IPAddress subnet(SUBNET_MASK);
+#else
+IPAddress subnet(0,0,0,0);
+#endif
+#ifdef GATEWAY_IP_ADDRESS
+IPAddress gateway(GATEWAY_IP_ADDRESS);
+#else
+IPAddress gateway(0,0,0,0);
+#endif
 
 int wifiConnectionAttemptCounter = 0;
 int wifiStatus = WL_IDLE_STATUS;
@@ -822,37 +832,37 @@ void systemResetCallback()
 }
 
 void printWifiStatus() {
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
   if ( WiFi.status() != WL_CONNECTED )
   {
     DEBUG_PRINT( "WiFi connection failed. Status value: " );
     DEBUG_PRINTLN( WiFi.status() );
   }
   else
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
   {
     // print the SSID of the network you're attached to:
     DEBUG_PRINT( "SSID: " );
 
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     DEBUG_PRINTLN( WiFi.SSID() );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
 
     // print your WiFi shield's IP address:
     DEBUG_PRINT( "IP Address: " );
 
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     IPAddress ip = WiFi.localIP();
     DEBUG_PRINTLN( ip );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
 
     // print the received signal strength:
     DEBUG_PRINT( "signal strength (RSSI): " );
 
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     long rssi = WiFi.RSSI();
     DEBUG_PRINT( rssi );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
 
     DEBUG_PRINTLN( " dBm" );
   }
@@ -873,6 +883,8 @@ void setup()
   DEBUG_PRINTLN( "using the WiFi 101 library." );
 #elif defined(ARDUINO_WIFI_SHIELD)
   DEBUG_PRINTLN( "using the legacy WiFi library." );
+#elif defined(ESP8266_WIFI)
+  DEBUG_PRINTLN( "using the ESP8266 WiFi library." );
 #elif defined(HUZZAH_WIFI)
   DEBUG_PRINTLN( "using the HUZZAH WiFi library." );
   //else should never happen here as error-checking in wifiConfig.h will catch this
@@ -886,7 +898,11 @@ void setup()
   DEBUG_PRINTLN( local_ip );
   //you can also provide a static IP in the begin() functions, but this simplifies
   //ifdef logic in this sketch due to support for all different encryption types.
+#ifdef ESP8266_WIFI  
+  stream.config( local_ip , gateway, subnet );
+#else  
   stream.config( local_ip );
+#endif  
 #else
   DEBUG_PRINTLN( "IP will be requested from DHCP ..." );
 #endif
--- a/examples/StandardFirmataWiFi/wifiConfig.h
+++ b/examples/StandardFirmataWiFi/wifiConfig.h
@@ -23,8 +23,9 @@
  */
 #define ARDUINO_WIFI_SHIELD
 
-//do not modify these next 4 lines
+//do not modify these next 5 lines
 #ifdef ARDUINO_WIFI_SHIELD
+#include <WiFi.h>
 #include "utility/WiFiStream.h"
 WiFiStream stream;
 #endif
@@ -44,16 +45,39 @@ WiFiStream stream;
  */
 //#define WIFI_101
 
-//do not modify these next 4 lines
+//do not modify these next 5 lines
 #ifdef WIFI_101
+#include <WiFi101.h>
 #include "utility/WiFi101Stream.h"
-WiFi101Stream stream;
+WiFiStream stream;
 #endif
 
 /*
- * OPTION C: Configure for HUZZAH
+ * OPTION C: Configure for ESP8266
  *
- * HUZZAH is not yet supported, this will be added in a later revision to StandardFirmataWiFi
+ * This will configure StandardFirmataWiFi to use the ESP8266WiFi library for boards
+ * with an ESP8266 chip. It is compatible with 802.11 B/G/N networks.
+ *
+ * To enable, uncomment the #define ESP8266_WIFI below and verify the #define values under
+ * options A and B are commented out.
+ *
+ * IMPORTANT: You must have the esp8266 board support installed. To easily install this board, open the board manager via:
+ * Arduino IDE Menus: Tools > Board > Manage Boards > filter search for "esp8266" > Select the result and click 'install'
+ */
+//#define ESP8266_WIFI
+
+//do not modify these next 5 lines
+#ifdef ESP8266_WIFI
+#include <ESP8266WiFi.h>
+#include "utility/WiFiStream.h"
+WiFiStream stream;
+#endif
+
+/*
+ * OPTION D: Configure for HUZZAH
+ *
+ * HUZZAH with CC3000 is not yet supported, this will be added in a later revision to StandardFirmataWiFi.
+ * For HUZZAH with ESP8266 use ESP8266_WIFI.
  */
 
 //------------------------------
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: examples/StandardFirmataWiFi/wifiConfig.h [hunk 1]
```diff
@@ -140,11 +153,11 @@ char wep_key[] = "your_wep_key";
  * CONFIGURATION ERROR CHECK (don't change anything here)
  *============================================================================*/
 
-#if ((defined(ARDUINO_WIFI_SHIELD) && (defined(WIFI_101) || defined(HUZZAH_WIFI))) || (defined(WIFI_101) && defined(HUZZAH_WIFI))  || (defined(WIFI_101) && defined(ESP_WIFI))  || (defined(ESP_WIFI) && defined(HUZZAH_WIFI)) || (defined(ESP_WIFI) && defined(ARDUINO_WIFI_SHIELD)))
+#if ((defined(ARDUINO_WIFI_SHIELD) && (defined(WIFI_101) || defined(HUZZAH_WIFI))) || (defined(WIFI_101) && defined(HUZZAH_WIFI))  || (defined(WIFI_101) && defined(ESP8266_WIFI))  || (defined(ESP8266_WIFI) && defined(HUZZAH_WIFI)) || (defined(ESP8266_WIFI) && defined(ARDUINO_WIFI_SHIELD)))
```
soundanalogous: we need to figure out a more scalable way to handle this

examples/StandardFirmataWiFi/wifiConfig.h
```diff
@@ -44,46 +45,57 @@ WiFiStream stream;
...
 // replace this with your wireless network SSID
 char ssid[] = "your_network_name";
 
 // STEP 3 [OPTIONAL for all boards and shields]
 // if you want to use a static IP (v4) address, uncomment the line below. You can also change the IP.
-// if this line is commented out, the WiFi shield will attempt to get an IP from the DHCP server
-// #define STATIC_IP_ADDRESS 192,168,1,113
+// if the first line is commented out, the WiFi shield will attempt to get an IP from the DHCP server
+#define STATIC_IP_ADDRESS  192,168,1,113
+#define SUBNET_MASK        255,255,255,0 // REQUIRED for ESP8266_WIFI, ignored for others
```
soundanalogous: Is subnet and gateway only required for ESP8266 if using Wi-Fi? If not, move these 2 defines into the  `#ifdef ESP8266_WIFI` block.

utility/WiFiStream.h
```diff
@@ -59,19 +60,37 @@ class WiFiStream : public Stream
 public:
   WiFiStream() {};
 
+#ifndef ESP8266
```
soundanalogous: `#ifndef ESP8266_WIFI`?

utility/WiFiStream.h
```diff
@@ -59,19 +60,37 @@ class WiFiStream : public Stream
...
+  // allows another way to configure a static IP before begin is called
+  inline void config(IPAddress local_ip, IPAddress gateway, IPAddress subnet)
+  {
+    _local_ip = local_ip;
+    _gateway = gateway;
+    _subnet = subnet;
+#ifdef ESP8266
+    WiFi.config( local_ip, gateway, subnet );
+#else
+    WiFi.config( local_ip, IPAddress(0, 0, 0, 0), gateway, subnet );
```
soundanalogous: when would this ever be used if gateway and subnet are specific to ESP8266?

utility/WiFiStream.h
```diff
@@ -124,6 +149,7 @@ class WiFiStream : public Stream
     return result;
   }
 
+#ifndef ESP8266
```
soundanalogous: is there a reason all of these begin functions need to be ignored for ESP8266?

soundanalogous:
This script can be used to test the Wi-Fi server implementation: https://gist.github.com/soundanalogous/2fbeb29c39f11498fd7f

I'll check with Arduino 101 shield and Arduino MKR1000 later today.

examples/StandardFirmataWiFi/wifiConfig.h
jnsbyr: For ARDUINO_WIFI_SHIELD and WIFI_101 one could provide the two include alternatives and use only one define.

I assume the HUZZAH_WIFI define is reserved for the CC3000 WiFi shield. I assume it comes with its own library, but I don't know if it is identical with WiFi/WiFi101. The HUZZAH ESP8266 is covered by the ESP8266.

With the ESP8266 one could use the architecture define already in use in Boards.h. This could replace the ESP_WIFI define and could automatically exclude all shields.

examples/StandardFirmataWiFi/wifiConfig.h
jnsbyr: Interestingly the WiFi and WiFi101 support local IP mode without the need to define netmask and gateway but they also support these parameters. From the network point of view the full set is preferable.

utility/WiFiStream.h
jnsbyr: Right, must be the other way round. The same applies to all the other ESP8266 "excludes" in this file.

utility/WiFiStream.h
jnsbyr: See my comment above to SUBNET_MASK and GATEWAY_IP_ADDRESS. I would prefer to always define them and to remove the variant with only a local IP but this needs testing and I don't have an Arduino with a WiFi shield.

utility/WiFiStream.h
jnsbyr: I could not find matching methods for WEP security in the ESP8266/arduino project. As far a I know the ESP8266 supports WEP. I will have another look into this. 

The other begin() functions also use the local IP only definition without netmask and gateway that is not supported by the ESP8266. All these functions are not used by the example.

jnsbyr: WEP is not mentioned [here](https://github.com/esp8266/Arduino/blob/master/doc/libraries.md#wifiesp8266wifi-library).

examples/StandardFirmataWiFi/wifiConfig.h
soundanalogous: one hacky way to do it, would be for each Wi-Fi driver definition, increment a counter (`wifiDriverCount`). Initialize it to 0. If no drivers will included `wifiDriverCount = 0`, throw an error. If more than 1 driver is included (`wifiDriverCount > 1`), throw an error.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
```diff
@@ -822,37 +832,37 @@ void systemResetCallback()
 }
 
 void printWifiStatus() {
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101)
+#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
```
soundanalogous: I don't think any of these #ifdefs are needed anymore since all supported libraries support the following function calls. If a lib does not support one of these calls in the future, we could use an #ifndef instead. I really want to try to reduce the number of conditional compiler statements in the StandardFirmata variants.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
```diff
@@ -125,6 +125,16 @@ SerialFirmata serialFeature;
 #ifdef STATIC_IP_ADDRESS
 IPAddress local_ip(STATIC_IP_ADDRESS);
 #endif
+#ifdef SUBNET_MASK
```
soundanalogous: maybe move all of this to wifiConfig.h?

soundanalogous:
I'm not having any luck with the Adafruit Huzzah ESP8266 board. I noticed there are some differences between the ESP8266 WiFi API and the Arduino WiFi api, such as `begin` does not return the state in the ESP8266 api ([so this will not return anything](https://github.com/jnsbyr/arduino/blob/47993810aa7b50e22c1ee810e2ad5f36bfaa9a74/utility/WiFiStream.h#L181)). I tried making some changes in WiFiStream such as using `WiFi.status()` to reflect this but no luck. Seems anything I try hangs the board while it's trying to connect via WPA.

soundanalogous:
I was able to get a WiFi client version with the ESP8266 simply using EthernetClientStream, inspired by this Fork: https://github.com/kotl/esp8266-firmata/commit/cfeb9c43c3dff236e229e049d79f78a8e22a9f7e.

Here is a test [sketch file](https://gist.github.com/soundanalogous/1d1458a16b64114899c24bf6cc8bf2da)
And the accompanying [config file](https://gist.github.com/soundanalogous/155ae5a1f9d3d1d72b8162cc3cbf5de7)

jnsbyr:
Using EthernetClientStream is what I referred to [here](https://github.com/firmata/arduino/pull/278#issuecomment-196056312). But you still need to use WiFi.begin() and WiFi.status() and this is also used in WiFiStream. I had no problem using WiFi.status() (see below). 

This is what I changed in StandardFirmataWiFi to make it work using ```HUZZAH_WIFI``` as switch and what I currently use with my Huzzah ESP8266. :

```C
#ifdef HUZZAH_WIFI
#include <ESP8266WiFi.h>
#include "utility/EthernetClientStream.h"
WiFiClient client;

#ifdef SERVER_IP
#ifdef STATIC_IP_ADDRESS
EthernetClientStream stream(client, IPAddress(STATIC_IP_ADDRESS), IPAddress(SERVER_IP), NULL, SERVER_PORT);
#else
EthernetClientStream stream(client, IPAddress(0, 0, 0, 0), IPAddress(SERVER_IP), NULL, SERVER_PORT);
#endif
#endif

...

#ifdef HUZZAH_WIFI  
  WiFi.config(IPAddress(STATIC_IP_ADDRESS), IPAddress(NETWORK), IPAddress(GATEWAY));
#else  
  stream.config(STATIC_IP_ADDRESS);
#endif

...

#elif defined(WIFI_WPA_SECURITY)
  while (wifiStatus != WL_CONNECTED) {
    DEBUG_PRINT( "Attempting to connect to WPA SSID: " );
    DEBUG_PRINTLN(ssid);
#if defined(HUZZAH_WIFI)    
    WiFi.begin(ssid, wpa_passphrase);
    wifiStatus = WiFi.status();
#else
    wifiStatus = stream.begin(ssid, wpa_passphrase, SERVER_PORT);
#endif        
    delay(5000); // TODO - determine minimum delay
    if (++wifiConnectionAttemptCounter > WIFI_MAX_CONN_ATTEMPTS) break;
  }

...

#if defined(HUZZAH_WIFI)    
  stream.maintain(IPAddress(STATIC_IP_ADDRESS));
#else
  stream.maintain();
#endif  
```

But this results in a lot of defines in the ino file. My idea is to use a base class WiFiStream with the common code for WiFi, client and server and a WiFiServerStream and a WiFiClientStream class that inherit from WiFiStream. This should reduce or at least hide the number of defines and would avoid mixing everything in one class with a lot of ifs for the client and server specials.

jnsbyr:
Had another look at kotl@cfeb9c4. Although not a default approach for Firmata but found in several IT devices the serial port is used to configure the network parameters. Another way to do it that is to define a protocol extension for the setup parameters and to start Firmata with invalid EEPROM data in server mode using DHCP. After configuration Firmata keeps its WiFi secrets and network parameters in EEPROM and may even switch to client mode. That could be an interesting project, but first things first: 

How should we proceed with the client/server integration? Separate INOs and/or separate classes with or without inheritance? As soon as that is decided, I could implement the client, test with the Huzzah ESP8266 and make the necessary changes for the server and the INO.

soundanalogous:
We could rename EthernetClientStream to NetworkClientStream since it works for both Ethernet and WiFi. However doing something similar for WiFiStream is tricky since I'd Ideally want an Ethernet server variant as well and unlike EthernetClientStream which makes no reference to the Ethernet or WiFi libraries, WiFiStream references the WiFi library.

Were you able to get WiFi server working on ESP8266 or only WiFi client?

soundanalogous:
This PR is also starting to get fragmented. It's partly about ESP8266 integration and partly about merging WiFiStream and WiFi101Stream and partly about creating a separate WiFi server and WiFi client implementation. Should probably be 3 different PRs or we'll end up with a rats nest.

soundanalogous:
I think we should proceed as follows:

1. Cherry pick commits that added ESP8266 to Boards.h to master.
2. Resolve WiFiStream and WiFi101Stream into a single WiFiStream class and merge to master
3. Figure out the best approach to having WiFi server and client support and work on this in a new branch
4. ESP8266 WiFi integration
5. Ethernet server

jnsbyr:
1. Go ahead by merging esp Boards.h into master.
2. This needs removing one include from WiFiStream.h, deleting WiFi101Stream and adding two includes to the wifiConfig.h.
3. Please create this branch, but I would prefer to have the ESP8266 along. Because of the slightly different requirements it could help find the best solution for the client/server support. Having ESP8266 on board from the beginning would also allow me to do tests with my hardware.

jnsbyr:
> Were you able to get WiFi server working on ESP8266 or only WiFi client?

I have the WiFiClient working in the configuration mentioned [above](https://github.com/firmata/arduino/pull/285#issuecomment-207950550), but I haven't tried anything with this PR or a WiFiServer so far.

I already considered reusing the EthernetClientStream because its working, but I still think WiFi.maintain() should be basically be the same for client and server so I suggest a WiFiStream/WiFiClientStream approach instead of sharing EthernetClientStream.

soundanalogous:
I have discovered what the issue is with ESP8266 WiFi server. In WiFiStream, [this line](https://github.com/firmata/arduino/blob/master/utility/WiFiStream.h#L148) returns 0 if a connection is established. I suspect this is to enable the user to establish a new connection with a different access point. This works fine with the Arduino WiFi library as [this line](https://github.com/firmata/arduino/blob/master/utility/WiFiStream.h#L154()) always seems to return the connected state with Arduino WiFi (I assume it waits) whereas it never does with ESP8266WiFi (I assume it does not wait). What happens then is with ESP8266 the connection status is updated later, but is never returned since a connected state upon calling `begin` always returns false.

I'm looking into solutions that don't involve adding additional #ifdefs to the sketch file but it has proven to be a bit tricky to find such a solution.

soundanalogous:
Maybe I could set a variable `is_new_connection` to true each time `WiFi.begin` is called and then change the following:

```
 if( !is_ready() ) return 0;
```

to something like:
```
if (is_new_connection && WiFi.status() == WL_CONNECTED) {
  is_new_connection = false;
  return WL_CONNECTED;
}
if (!is_ready()) return 0;
```

Something like that would be more stable for both ESP8266 and Arduino WiFi implementations.

jnsbyr:
> I assume it does not wait

That is very probable. With the ESP8266 waiting is not a good idea. After a max. of 1 second the watchdog will trigger. Everything that intends to wait must periodically exit the main loop or call yield().

From my point of view a WiFi TCP connections differs from Ethernet TCP connection becaus it needs two connects. First there is the connection from the WiFi station to the AP and then there is the connection from the TCP client to the server. They should be handled separately and hierarchically. As long as there is no WiFi link one does not need to bother with the TCP link. Maybe using WiFi.status() is the common ground of Arduino and ESP8266. Establishing the WiFi link and the TCP connection should be performed non-blocking if possible, at least when reconnecting, better always. 

Let begin() just do the initialisation and then maintain() can do the rest.

Issue #280, opened 2016-03-29
no version without query with Ethernet Firmata


jnsbyr:
```Firmata.begin``` pushes the protocol and firmware version on the Stream regardless of its connected state. With USB powered devices using a serial host link this is no problem because booting Firmata and registering the USB driver on the host takes approximately the same time so the serial host typically receives the protocol and firmware versions without asking. With StandardFirmataEthernet this data gets lost because there is no connection when the data is written and you need to query the device explicitly for the versions. 

At least with ```EthernetClient``` this can easily be amended by a simple change in the ino file by waiting for the initial connect:
```C
void setup()
{
  ...
  // start up Network Firmata:
  while (!client.connected()) { // wait for connect to avoid loosing initial messages
    stream.flush(); 
  };  
  Firmata.begin(stream);
  systemResetCallback();  // reset to default config
}
```
This is a solution as long as the Firmata firmware has no custom code that needs to run regardless of the host connect. 

Another way to do this would be a new "hostConnectedCallback" that could handle this regardless of the host connection type. 

But with TCP Ethernet connections there is another aspect to consider: TCP connection can fail and can be reestablished completely independent of the Firmata device state. Some hosts would like to know if the Firmata device has been restarted or if the TCP connection has been reestablished. For this one could skip the automatic version transmit on reconnect or send a new telegram telling the host that this was just a reconnect.

soundanalogous:
The reason there is typically no issue with USB is that when a client connects via serial, the board (unless it's an ATmega32u4 based board) hard resets.

A client can send a version query at any time so I'm not sure that waiting to transmit the version until a non-serial transport is established is all that important. However the ability to know if a client disconnected and then reconnected would be useful and has been requested several times in the past.

A "hostConnectedCallback" as you suggest sounds like an interesting approach. Do you want to draft something along those lines?

soundanalogous:
Keep in mind there are now also WiFi and BLE transports for Firmata as well.

jnsbyr:
Thanks for your feedback. As far as I can see most Arduino communication libraries do not notify so a hostConnectedCallback can only be done by polling ```connected``` for a state change. I will have a look at the BLE API because I haven't used it so far before making more detailed suggestions.

soundanalogous:
At some point we'll be able to leverage the `availableForWrite` method once it's added to Ethernet (and hopefully other Stream client classes).

[...]

PR #289, opened 2016-04-15
WiFiClientStream added

jnsbyr:
Changes
- server related functions moved from WiFiStream into WiFiServerStream
- unified class interface for sketch file (no client/server defines)
- WiFi and TCP configuration and connect split into individual methods (TCP constructor, static IP config, WiFi begin and TCP maintain)
- WiFi only initialized once at startup (WiFi connections cannot be forced by repeatedly calling begin)
- convenience methods for combined WiFi/network initialization removed

Tests
- tested with Huzzah ESP8266 in client mode
- successfully builds for AVR architecture with legacy WiFi option

Todo
- decide if inheritance from new WiFiStream is preferable to copied code in two classes in regard to code efficiency, code size, platform compatibility and maintenance
- decide if a single client/server class would be preferable to separate classes in regard to code efficiency, code size and maintenance

The approach used in this PR could be applied to EthernetClientStream to provide the missing EthernetServerStream. Basically its the same without the WiFi parts.

### DIFF
--- a/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
+++ b/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
@@ -918,6 +918,7 @@ void setup()
    */
   while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
     delay(500);
+    DEBUG_PRINT(".");
     streamConnected = stream.maintain();
   }
   if (streamConnected) {  
--- a/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
+++ b/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
@@ -110,7 +110,7 @@
 // the minimum interval for sampling analog input
 #define MINIMUM_SAMPLING_INTERVAL   1
 
-#define WIFI_MAX_CONN_ATTEMPTS      3
+#define MAX_CONN_ATTEMPTS           20  // [500 ms] -> 10 s
 
 /*==============================================================================
  * GLOBAL VARIABLES
@@ -130,8 +130,8 @@ IPAddress subnet(SUBNET_MASK);
 IPAddress gateway(GATEWAY_IP_ADDRESS);
 #endif
 
-int wifiConnectionAttemptCounter = 0;
-int wifiStatus = WL_IDLE_STATUS;
+int connectionAttempts = 0;
+bool streamConnected = false;
 
 /* analog inputs */
 int analogInputsToReport = 0;      // bitwise array to store pin reporting
@@ -307,6 +307,12 @@ void checkDigitalInputs(void)
   if (TOTAL_PORTS > 14 && reportPINs[14]) outputPort(14, readPort(14, portConfigInputs[14]), false);
   if (TOTAL_PORTS > 15 && reportPINs[15]) outputPort(15, readPort(15, portConfigInputs[15]), false);
 }
+
+// -----------------------------------------------------------------------------
+// function forward declarations
+void enableI2CPins();
+void disableI2CPins();
+void reportAnalogCallback(byte analogPin, int value);
 
 // -----------------------------------------------------------------------------
 /* sets the pin mode to the correct state and sets the relevant bits in the
@@ -826,38 +832,26 @@ void systemResetCallback()
 }
 
 void printWifiStatus() {
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
   if ( WiFi.status() != WL_CONNECTED )
   {
     DEBUG_PRINT( "WiFi connection failed. Status value: " );
     DEBUG_PRINTLN( WiFi.status() );
   }
   else
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
   {
     // print the SSID of the network you're attached to:
     DEBUG_PRINT( "SSID: " );
-
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     DEBUG_PRINTLN( WiFi.SSID() );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
 
     // print your WiFi shield's IP address:
     DEBUG_PRINT( "IP Address: " );
-
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     IPAddress ip = WiFi.localIP();
     DEBUG_PRINTLN( ip );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
 
     // print the received signal strength:
     DEBUG_PRINT( "signal strength (RSSI): " );
-
-#if defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
     long rssi = WiFi.RSSI();
     DEBUG_PRINT( rssi );
-#endif    //defined(ARDUINO_WIFI_SHIELD) || defined(WIFI_101) || defined(ESP8266_WIFI)
-
     DEBUG_PRINTLN( " dBm" );
   }
 }
@@ -890,7 +884,7 @@ void setup()
 #ifdef STATIC_IP_ADDRESS
   DEBUG_PRINT( "Using static IP: " );
   DEBUG_PRINTLN( local_ip );
-#ifdef ESP8266_WIFI
+#if defined(ESP8266_WIFI) || (defined(SUBNET_MASK) && defined(GATEWAY_IP_ADDRESS))
   stream.config( local_ip , gateway, subnet );
 #else
   // you can also provide a static IP in the begin() functions, but this simplifies
@@ -902,37 +896,35 @@ void setup()
 #endif
 
   /*
-   * Configure WiFi security
+   * Configure WiFi security and initiate WiFi connection
    */
 #if defined(WIFI_WEP_SECURITY)
-  while (wifiStatus != WL_CONNECTED) {
     DEBUG_PRINT( "Attempting to connect to WEP SSID: " );
     DEBUG_PRINTLN(ssid);
-    wifiStatus = stream.begin( ssid, wep_index, wep_key, SERVER_PORT );
-    delay(5000); // TODO - determine minimum delay
-    if (++wifiConnectionAttemptCounter > WIFI_MAX_CONN_ATTEMPTS) break;
-  }
-
+  stream.begin(ssid, wep_index, wep_key);
 #elif defined(WIFI_WPA_SECURITY)
-  while (wifiStatus != WL_CONNECTED) {
     DEBUG_PRINT( "Attempting to connect to WPA SSID: " );
     DEBUG_PRINTLN(ssid);
-    wifiStatus = stream.begin(ssid, wpa_passphrase, SERVER_PORT);
-    delay(5000); // TODO - determine minimum delay
-    if (++wifiConnectionAttemptCounter > WIFI_MAX_CONN_ATTEMPTS) break;
-  }
-
+  stream.begin(ssid, wpa_passphrase);
 #else                          //OPEN network
-  while (wifiStatus != WL_CONNECTED) {
     DEBUG_PRINTLN( "Attempting to connect to open SSID: " );
     DEBUG_PRINTLN(ssid);
-    wifiStatus = stream.begin(ssid, SERVER_PORT);
-    delay(5000); // TODO - determine minimum delay
-    if (++wifiConnectionAttemptCounter > WIFI_MAX_CONN_ATTEMPTS) break;
-  }
+  stream.begin(ssid);
 #endif //defined(WIFI_WEP_SECURITY)
-
   DEBUG_PRINTLN( "WiFi setup done" );
+
+  /*
+   * Wait for TCP connection to be established
+   */
+  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
+    delay(500);
+    streamConnected = stream.maintain();
+  }
+  if (streamConnected) {  
+    DEBUG_PRINTLN( "TCP connection established" );
+  } else {
+    DEBUG_PRINTLN( "failed to establish TCP connection" );    
+  }
   printWifiStatus();
 
   /*
--- a/examples/StandardFirmataWiFi/wifiConfig.h
+++ b/examples/StandardFirmataWiFi/wifiConfig.h
@@ -30,16 +30,16 @@
  */
 //#define WIFI_101
 
-//do not modify the following 10 lines
+//do not modify the following 11 lines
 #if defined(ARDUINO_SAMD_MKR1000) && !defined(WIFI_101)
 // automatically include if compiling for MRK1000
 #define WIFI_101
 #endif
 #ifdef WIFI_101
 #include <WiFi101.h>
-#include "utility/WiFiStream.h"
-WiFiStream stream;
-#define WIFI_LIB_INCLUDED
+#include "utility/WiFiClientStream.h"
+#include "utility/WiFiServerStream.h"
+  #define WIFI_LIB_INCLUDED
 #endif
 
 /*
@@ -57,8 +57,8 @@ WiFiStream stream;
 //do not modify the following 10 lines
 #ifdef ARDUINO_WIFI_SHIELD
 #include <WiFi.h>
-#include "utility/WiFiStream.h"
-WiFiStream stream;
+#include "utility/WiFiClientStream.h"
+#include "utility/WiFiServerStream.h"
   #ifdef WIFI_LIB_INCLUDED
   #define MULTIPLE_WIFI_LIB_INCLUDES
   #else
@@ -85,8 +85,8 @@ WiFiStream stream;
 #endif
 #ifdef ESP8266_WIFI
 #include <ESP8266WiFi.h>
-#include "utility/WiFiStream.h"
-WiFiStream stream;
+#include "utility/WiFiClientStream.h"
+#include "utility/WiFiServerStream.h"
   #ifdef WIFI_LIB_INCLUDED
   #define MULTIPLE_WIFI_LIB_INCLUDES
   #else
@@ -122,12 +122,17 @@ char ssid[] = "your_network_name";
 //#define GATEWAY_IP_ADDRESS 0,0,0,0       // REQUIRED for ESP8266_WIFI, optional for others
 
 
-// STEP 4 [REQUIRED for all boards and shields]
+// STEP 4 [OPTIONAL for all boards and shields]
+// uncomment and replace with the IP address of your server if the Arduino is the TCP client
+//#define SERVER_IP 10, 0, 0, 15
+
+
+// STEP 5 [REQUIRED for all boards and shields]
 // define your port number here, you will need this to open a TCP connection to your Arduino
 #define SERVER_PORT 3030
 
 
-// STEP 5 [REQUIRED for all boards and shields]
+// STEP 6 [REQUIRED for all boards and shields]
 // determine your network security type (OPTION A, B, or C). Option A is the most common, and the
 // default.
 
@@ -200,6 +205,16 @@ char wep_key[] = "your_wep_key";
 #error "you must choose between WIFI_NO_SECURITY and WIFI_WPA_SECURITY"
 #endif
 
+/*==============================================================================
+ * WIFI STREAM (don't change anything here)
+ *============================================================================*/
+
+#ifdef SERVER_IP
+  WiFiClientStream stream(IPAddress(SERVER_IP), SERVER_PORT);
+#else
+  WiFiServerStream stream(SERVER_PORT);
+#endif
+ 
 /*==============================================================================
  * PIN IGNORE MACROS (don't change anything here)
  *============================================================================*/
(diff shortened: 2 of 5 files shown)


examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
```diff
@@ -902,37 +896,35 @@ void setup()
...
+   * Wait for TCP connection to be established
+   */
+  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
+    delay(500);
+    streamConnected = stream.maintain();
+  }
+  if (streamConnected) {  
+    DEBUG_PRINTLN( "TCP connection established" );
+  } else {
+    DEBUG_PRINTLN( "failed to establish TCP connection" );
```
soundanalogous: This is also be reported in `printWiFiStatus()`. Do we really need it in both places?

examples/StandardFirmataWiFi/wifiConfig.h
```diff
@@ -30,16 +30,16 @@
...
 #if defined(ARDUINO_SAMD_MKR1000) && !defined(WIFI_101)
 // automatically include if compiling for MRK1000
 #define WIFI_101
 #endif
 #ifdef WIFI_101
 #include <WiFi101.h>
-#include "utility/WiFiStream.h"
-WiFiStream stream;
-#define WIFI_LIB_INCLUDED
+#include "utility/WiFiClientStream.h"
```
soundanalogous: What is the memory overhead of including both client and server streams simultaneously?

soundanalogous:
This looks awesome! You've been busy :)
Will test with WiFi 101 shield and MKR1000.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
```diff
@@ -309,6 +309,12 @@ void checkDigitalInputs(void)
 }
 
 // -----------------------------------------------------------------------------
+// function forward declarations
+void enableI2CPins();
```
soundanalogous: which compiler needs these?

soundanalogous:
WiFi client works for ESP8266, MKR1000 and WiFi 101 shield with Arduino Zero. I see you made a change however in that at startup it waits for a client connection instead of waiting for the router connection. This makes sense, but I wonder if instead of having a max attempts, it should just loop until a connection is made and the debug messages are updated accordingly. Otherwise it will appear to the user that the connection has failed.

Server mode does not work. The connection exceeds max attempts, however an IP address is assigned but I cannot establish a connection with the TCP client application.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
```diff
@@ -902,37 +896,35 @@ void setup()
...
 #endif //defined(WIFI_WEP_SECURITY)
-
   DEBUG_PRINTLN( "WiFi setup done" );
+
+  /*
+   * Wait for TCP connection to be established
+   */
+  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
+    delay(500);
+    streamConnected = stream.maintain();
```
soundanalogous: add: `DEBUG_PRINT(".");` after the delay as a progress indicator

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
soundanalogous: This is also misleading currently as I get the failed message if I am not fast enough in starting up my host application. Maybe the while loop should continue until a connection is made and/or use a much longer timeout). Previously the while loop only waited until a connection was made with the router.

soundanalogous:
Tried digging into WiFi server issues. No luck. I don't get any errors so I'm not sure where the connection is failing.

soundanalogous:
Working backwards from the old WiFiStream to something closer to what you have, but without inheriting from WiFiStream yet, the following works for the server implementation. The differences are subtle from what you have (minus the subclassing). I added `_server = WiFiServer(_port);` to the `maintain` method.

```c++
#ifndef WIFI_SERVER_STREAM_H
#define WIFI_SERVER_STREAM_H

#include <inttypes.h>
#include <Stream.h>

class WiFiServerStream : public Stream
{
private:
  WiFiServer _server = WiFiServer(23);
  WiFiClient _client;

  //configuration members
  IPAddress _local_ip;
  IPAddress _gateway;
  IPAddress _subnet;
  uint16_t _port = 0;
  uint8_t _key_idx = 0;               //WEP
  const char *_key = nullptr;         //WEP
  const char *_passphrase = nullptr;  //WPA
  char *_ssid = nullptr;
  bool _listening = false;

  inline bool connect_client()
  {
    if(_client && _client.connected()) return true;

    WiFiClient newClient = _server.available();
    if( !newClient ) return false;
    _client = newClient;

    return true;
  }

public:
  WiFiServerStream(uint16_t server_port) : _port(server_port) {}

#ifndef ESP8266
  // allows another way to configure a static IP before begin is called
  inline void config(IPAddress local_ip)
  {
    _local_ip = local_ip;
    WiFi.config( local_ip );
  }
#endif

  // allows another way to configure a static IP before begin is called
  inline void config(IPAddress local_ip, IPAddress gateway, IPAddress subnet)
  {
    _local_ip = local_ip;
    _gateway = gateway;
    _subnet = subnet;
#ifndef ESP8266
    WiFi.config( local_ip, IPAddress(0, 0, 0, 0), gateway, subnet );
#else
    WiFi.config( local_ip, gateway, subnet );
#endif
  }

  // get DCHP IP
  inline IPAddress getLocalIP()
  {
    return WiFi.localIP();
  }

  /**
   * @return true if connected
   */
  inline bool maintain()
  {
    if( connect_client() ) return true;

    stop();

    if( !_listening && WiFi.status() == WL_CONNECTED )
    {
      // start TCP server after first WiFi connect
      _server = WiFiServer(_port);
      _server.begin();
      _listening = true;
    }

    return false;
  }

/******************************************************************************
 *           Connection functions with DHCP
 ******************************************************************************/

  //OPEN networks
  inline int begin(char *ssid)
  {
    _ssid = ssid;

    WiFi.begin(ssid);
    return WiFi.status();
  }

#ifndef ESP8266
  //WEP-encrypted networks
  inline int begin(char *ssid, uint8_t key_idx, const char *key)
  {
    _ssid = ssid;
    _key_idx = key_idx;
    _key = key;

    WiFi.begin(ssid, key_idx, key);
    return WiFi.status();
  }
#endif

  //WPA-encrypted networks
  inline int begin(char *ssid, const char *passphrase)
  {
    _ssid = ssid;
    _passphrase = passphrase;

    WiFi.begin(ssid, passphrase);
    return WiFi.status();
  }

/******************************************************************************
 *           Connection functions without DHCP
 ******************************************************************************/

// ESP8266 requires gateway and subnet so the following functions are not compatible
#ifndef ESP8266
  //OPEN networks with static IP
  inline int begin(char *ssid, IPAddress local_ip)
  {
    _ssid = ssid;
    _local_ip = local_ip;

    WiFi.config( local_ip );
    WiFi.begin(ssid);
    return WiFi.status();
  }

  //WEP-encrypted networks with static IP
  inline int begin(char *ssid, IPAddress local_ip, uint8_t key_idx, const char *key)
  {
    _ssid = ssid;
    _local_ip = local_ip;
    _key_idx = key_idx;
    _key = key;

    WiFi.config( local_ip );
    WiFi.begin(ssid, key_idx, key);
    return WiFi.status();
  }

  //WPA-encrypted networks with static IP
  inline int begin(char *ssid, IPAddress local_ip, const char *passphrase)
  {
    _ssid = ssid;
    _local_ip = local_ip;
    _passphrase = passphrase;

    WiFi.config( local_ip );
    WiFi.begin(ssid, passphrase);
    return WiFi.status();
  }
#endif

/******************************************************************************
 *             Stream implementations
 ******************************************************************************/

  inline int available()
  {
    return connect_client() ? _client.available() : 0;
  }

  inline void flush()
  {
    if( _client ) _client.flush();
  }

  inline int peek()
  {
    return connect_client() ? _client.peek(): 0;
  }

  inline int read()
  {
    return connect_client() ? _client.read() : -1;
  }

  inline void stop()
  {
    _client.stop();
  }

  inline size_t write(uint8_t byte)
  {
    if( connect_client() ) _client.write( byte );
  }
};

#endif //WIFI_SERVER_STREAM_H
```

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
jnsbyr: I didn't have a looked at the debug output so far.

Waiting until connected is a drawback if you add active logic that runs independently of the host. That is one reason why I wanted to introduce a [hostConnectedCallback](https://github.com/firmata/arduino/issues/280#issuecomment-204777188). 

Regarding the timeout, I reduced it by about 30% because an ESP8266 with static IP typically connects in less than 1.5 seconds. Raising the timeout should help somewhat.

examples/StandardFirmataWiFi/wifiConfig.h
jnsbyr: Good question. Commenting out one include and looking at the build size should tell. A good linker should notice that the code is not used and skip it.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
jnsbyr: The Xtensa Compiler for the ESP8266 can't do without forward declarations if you want to call a function before it is defined. I havn't checked the dependencies. If there are no circular references  reordering the functions can avoid this.

jnsbyr:
Interesting. I would have prefered to do the sever tests myself but didn't find the time to setup something like you already suggested so far. 

I wondered how the WiFi connection is triggered in server mode because there is no activity that initiates outgoing connections. Reopening the server port again and again will probably do the same as the client connection. This is the approach of the original WiFiStream. I will put the ```WiFiServer(_port)``` into maintain.

examples/StandardFirmataWiFi/wifiConfig.h
jnsbyr: With the Xtensa compiler there is no difference.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
jnsbyr:
```C
#define MAX_CONN_ATTEMPTS           20  // [500 ms] -> 10 s
```
should be long enough for a WiFi connect + DHCP + TCP connect in most cases, but I know there are situations where it needs more time, especially if the signal strength is weak. I am still not sure if this is reason enough to use a higher timeout. An easy way to cope with different requirements would be to move the timeout value into the config file and assign a default.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
jnsbyr: After reordering the functions in the sketch the forward definitions can be removed. But there is also the question of context, of keeping certain functions in a textual proximity or order to each other.

jnsbyr:
Had a closer look at what happens if you place ```WiFiServer(_port)``` into ```maintain()```. This is probably the TCP bind of the server port to the WiFi interface. If one does this already in the constructor, there is probably no interface (or at least no IP) to bind to. When waiting for the WiFi connect in ```maintain()``` at least the interface exists. When using DHCP it depends on timing if the IP is assigned before the ```WiFiServer(_port)``` in ```maintain()``` or not but I am not sure if the IP assignment is a requirement, because default port binding does not need an IP address.

utility/WiFiServerStream.h
```diff
@@ -0,0 +1,99 @@
...
+    if( _client && _client.connected() ) return true;
+
+    if( _connected )
+    {
+      stop();
+    }
+
+    // passive TCP connect (accept)
+    WiFiClient newClient = _server.available();
+    if( !_client ) return false;  // could this work on all platforms? if( !(_client && _client.connected()) ) return false;
```
soundanalogous: This is the bug that is preventing WiFi server from working.
change `!_client` to `!newClient`, otherwise the client was never being set.

utility/WiFiServerStream.h
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * check if TCP client is connected
+   * @return true if connected
+   */
+  virtual inline bool connect_client()
+  {
+    if( _client && _client.connected() ) return true;
+
+    if( _connected )
```
soundanalogous: `_connected` is never set to `true` so this block should either be removed or `_connected` should be set to `true` somewhere.

utility/WiFiServerStream.h
```diff
@@ -70,14 +70,15 @@ class WiFiServerStream : public WiFiStream
     if( connect_client() ) return true;
 
     stop();
-    
+
     if( !_listening && WiFi.status() == WL_CONNECTED )
     {
       // start TCP server after first WiFi connect
+      _server = WiFiServer(_port);
```
soundanalogous: It turns out that this change was not necessary (I check to be sure). The real issue was that the `_client` reference was never set.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
soundanalogous: Strange I have not run into any compiler errors when using the ESP8266 Arduino core and I have successfully compiled each module that library supports (except ThaiEasyElec's ESPino which fails). At least not as of version 2.1.0.

jnsbyr: While compiling I see the version 1.20.0-26 for the Xtensa and a lot of compiler options that are handled by the Arduino IDE. Maybe one of them makes the difference (e.g. for Firmata.cpp):

```
-U__STRICT_ANSI__ -c  -Os -g -mlongcalls -mtext-section-literals -fno-exceptions -fno-rtti -falign-functions=4 -std=c++11 -MMD -ffunction-sections -fdata-sections -DF_CPU=80000000L   -DARDUINO=10607 -DARDUINO_ESP8266_ESP12 -DARDUINO_ARCH_ESP8266  -DESP8266
```

utility/WiFiServerStream.h
jnsbyr: Great, you found it. Thx for testing. Have pushed the fixed version.

utility/WiFiServerStream.h
jnsbyr: Now this is also fixed in the new version. 

Added the host connection callback because updating _connected and notifying the callback must happen at the same time anyway. I didn't make any changes to the sketch file in regard to the host connected callback though.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
soundanalogous: I see, you are compiling via the command line rather than using the Arduino IDE.

examples/StandardFirmataWiFi/wifiConfig.h
soundanalogous: Verified there is also no difference when compiling for MKR1000 or Arduino Zero (with WiFi 101 shield)

soundanalogous:
I've tested both client and server on all of my WiFi boards and shields and both are working well.

However this block doesn't make sense when using as a TCP Server since a client application could connect at any time, even hours or days later vs running the board as a TCP client where you expect server application to already be running on the host on startup:

```c++
  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
    delay(500);
    DEBUG_PRINT(".");
    streamConnected = stream.maintain();
  }
  if (streamConnected) {  
    DEBUG_PRINTLN( "TCP connection established" );
  } else {
    DEBUG_PRINTLN( "failed to establish TCP connection" );    
  }
```

I think we may need to do something like this since when running as a TCP server we really want to know if the server is ready rather than if a client happens to connect within the first 10 seconds:

```c++
#ifdef SERVER_IP
  while (WiFi.status() != WL_CONNECTED && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
    delay(500);
    DEBUG_PRINT(".");
  }
#else
  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
    delay(500);
    DEBUG_PRINT(".");
    streamConnected = stream.maintain();
  }
  if (streamConnected) {  
    DEBUG_PRINTLN( "TCP connection established" );
  } else {
    DEBUG_PRINTLN( "failed to establish TCP connection" );    
  }
#endif
```

soundanalogous:
correction (erroneously had `#ifdef SERVER_IP` previously):
```c++
#ifndef SERVER_IP
  while (WiFi.status() != WL_CONNECTED && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
    delay(500);
    DEBUG_PRINT(".");
  }
#else
  while (!streamConnected && ++connectionAttempts <= MAX_CONN_ATTEMPTS) {
    delay(500);
    DEBUG_PRINT(".");
    streamConnected = stream.maintain();
  }
  if (streamConnected) {  
    DEBUG_PRINTLN( "TCP connection established" );
  } else {
    DEBUG_PRINTLN( "failed to establish TCP connection" );    
  }
#endif
```

Also, the WiFi101 library seems to block on WiFi.begin until a connection state change is observed, but ESP8266 does not so the loop above would be required. Otherwise I would suggest simply skipping the `streamConnected` loop for TCP server.

examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
jnsbyr: No, I just changed to verbosity via IDE settings to see what is really happening.

jnsbyr:
What you suggest would work, but I think the right way to do it is to use the host connection callback. 

One should not need to care if one uses a client or a sever. All that counts is that the TCP connection is established - and why wait at all? Just move most of the code from ```setup()``` after the ```stream.begin()``` call into the host connection callback.

soundanalogous:
There is a lot of one time init code in setup that has nothing to do with the transport that wouldn't make sense to run a second time. I was thinking the host connection callback controls whether or not the code in the loop function runs or is skipped entirely and the buffer is appropriately managed.

jnsbyr:
The host connection callback notifies when the TCP link state changes between connected and not connected. Its up to the sketch what to do with these events. I suggest splitting the current ```setup()``` into at least 2 parts, one that initializes the board after boot in ```setup()``` as it does  and one that needs to run after the initial host connect using the host connection callback. 

One could easily add a third state to the host connection callback handling, so that you get connected, disconnected and reconnected. But maybe it is better to let the sketch decide on the application layer if the initial setup by the host was really completed so that the next connect is treated as a reconnect. If you use this flag in the connected event you can decide when to do a special setup with the 1st host connect.

I would typically not skip any logic processing in the main loop while not connected. The Firmata transmit buffers is something different though.

soundanalogous:
I think the following should only be called once (if connect is a one time event - vs reconnect than that is okay as well).

- Firmata.setFirmwareVersion(...);
- All of the the pin ignore code
- Firmata.begin(stream);
- systemResetCallback(); // since you may not want to reinitialize when temporarily losing connection
- printWiFiStatus();

You'd have to block loop completely until the setup code runs.

It may make sense to attache and detach callbacks on each disconnect / reconnect. I'm not sure if this would cause any memory leaks or not.

The stream connection status could be reported on connect, disconnect and reconnect.

soundanalogous:
> You'd have to block loop completely until the setup code runs.
Or at least block `Firmata.processInput` if the callbacks are not attached because `processInput` is the code in Firmata.cpp that calls the callback functions.

soundanalogous:
Can you draft a version with the callbacks or would you rather do that in a separate PR?

jnsbyr:
It's hard for me to catch up right now. 

I agree that the things you listed above should be one timers that should stay in ```setup()```. But why do you worry about ```Firmata.processInput```? The WiFiStream abstraction will return _-1_ while not connected and nothing will happen and when the connection event occurs the host connection callback will be the first to be called.

The callback is already included in the current WiFiStream. If you want to give it a try add ```stream.attach(hostConnectionCallback)``` somewhere before calling ```stream.begin()``` and add your callback that might start like that:

```C
void hostConnectionCallback(byte state)
{
  byte port, portConfigDigital, pin, pinMode, analogPin;
  
  switch (state) {
    case HOST_CONNECTION_CONNECTED:
....
  }
}
```

The details of the host connection callback implementation in the sketch file are good for a separate PR. I leave it up to you if you want to initiate this PR yourself because for moving code in the sketch file there are things to consider that you are more aware of than me.

soundanalogous:
Okay I will merge this into the main wifi-client branch since I think it's in a good state and then I'll work on adding the connection host callback to StandardFirmataWiFi in a separate PR.

soundanalogous:
It's now clear to me what the issue is with deferred initialization after having a go at it. The problem is that until `Firmata.begin(stream)` is called, the `FirmataStream` pointer in Firmata.cpp remains uninitialized. This means when `loop` runs and any of the stream methods are called, there is no actual stream object to reference so the sketch will crash. Once the stream is initialized however, as you pointed out the `connect_client()` call in each of the stream methods ensures the client buffer is neither read nor written to so it should be safe even if there is no easy way to defer Firmata setup until after the TCP connection is established. In this case the host connection callback is probably most useful for indicating state to the user (if DEBUG is enabled) and managing the buffer or flags that may be useful).

soundanalogous:
The host connection callback is never called when using the ESP8266 as a TCP server. It is called however when using the WiFi101 library. When using the ESP8266 as a TCP client, host connection callback is only called for the CONNECTED state and only ever once. It's called for both states when using WiFi101 as a TCP client and called each time the client disconnect and reconnects to the server.

soundanalogous:
Here are the changes I made: https://github.com/firmata/arduino/pull/290

jnsbyr:
> The host connection callback is never called when using ...

I would not have expected the platforms to be that different. But there is probably a way to do detect a connect and a disconnect on all platforms. Will have a look at the ESP8266 specifics but I am not sure if I can provide any answers until the weekend.

soundanalogous:
No problem. I typically only have time to work on Firmata on the weekend as well.

PR #291, opened 2016-04-23
connection state management modified

jnsbyr:
- check member _connected first in connect_client() to prevent querying
connected() of disconnected _client
- added method status() to check TCP connection state (ESP8266 only)

### DIFF
--- a/utility/WiFiClientStream.h
+++ b/utility/WiFiClientStream.h
@@ -20,7 +20,7 @@
 
   published under the same license.
 
-  Last updated April 17th, 2016
+  Last updated April 23rd, 2016
  */
 
 #ifndef WIFI_CLIENT_STREAM_H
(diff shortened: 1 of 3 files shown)


jnsbyr:
This change may help to resolve the cross platform connect/disconnect issues mentioned [here](https://github.com/firmata/arduino/pull/289#issuecomment-211130369).

For the ESP8266 an additional change is required that is not included in this merge request. At least when SERIAL_DEBUG is enabled, calling ```pinMode(1, OUTPUT);``` in ```systemResetCallback()``` causes a  hang/crash at startup with 1 being the serial TX pin already in use. A possible solution could be:

```C
  for (byte i = 0; i < TOTAL_PINS; i++) {
    // pins with analog capability default to analog input
    // otherwise, pins default to digital output
    if (IS_PIN_ANALOG(i)) {
      // turns off pullup, configures everything
      setPinModeCallback(i, PIN_MODE_ANALOG);
    } else if (IS_PIN_DIGITAL(i)) {
#if defined(ESP8266)      
      if (i == 1) continue;
#endif      
      // sets the output to 0, configures portConfigInputs
      setPinModeCallback(i, OUTPUT);
    }
```
With these two changes combined connect and disconnect is signalled each time in client and server mode on the  ESP8266.

soundanalogous:
This should probably use a pin ignore macro:

`#define IS_IGNORE_ESP8266(p)     ((p) == 1 && SERIAL_DEBUG)`

I can add this later since I actually want to simplify this to use a single `IS_IGNORE_PIN` macro rather than being architecture specific which will remove a few more #ifdefs from StandardFirmataWiFi.

jnsbyr:
Made another test that confirms that the issue with pin 1 only arises when SERIAL_DEBUG is enabled.

Another  temporary way to avoid the pin 1 issue could be to add ```Firmata.setPinMode(1, PIN_MODE_IGNORE);``` to ```ignoreWiFiPins()``` for the ESP8266 with SERIAL_DEBUG.

The ```IS_IGNORE_PIN``` approach that you suggest is the better solution though.

soundanalogous:
Tested in both client and server modes with all of my Wi-Fi boards and shields and everything is working as expected. Merging...

PR #292, opened 2016-04-24
Add TCP client and ability to choose between TCP client and Server

soundanalogous:
Adds ability to setup board as a TCP client or TCP server.

### DIFF
--- a/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
+++ b/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
@@ -110,7 +110,7 @@
 // the minimum interval for sampling analog input
 #define MINIMUM_SAMPLING_INTERVAL   1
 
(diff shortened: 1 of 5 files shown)


soundanalogous:
@jnsbyr I think this is in a good state to merge into master. Let me know if you agree and/or if you had anything else to add.

Next step is to finalize the protocol definitions for state management (connect/reconnect and watchdog) and pin value reporting.

jnsbyr:
Had a look at the diffs and they look pretty good to me. You already tested several possible configurations so there is nothing I could add at this point.

It's agreed that we should continue with the protocol definitions. Maybe it is a good idea to create a watchdog test case both for a TCP and serial host link before finalizing the specifications.

soundanalogous:
I hadn't thought of the watchdog being necessary for serial since it's less likely in my experience at least to lose the connection, but I guess it could be useful for long running processes where power could be lost for some period of time causing a lose in the serial connection.

PR #297, opened 2016-07-30
Removed some spelling mistakes in the comments

shantanu69:
In line 6 the `clink` should be click and in line 7 there should be a `in` before `your default browser`


soundanalogous:
Thanks for catching this. Could you make the same change in all of the StandardFirmata... files?

shantanu69:
Okay, I will see :+1:

shantanu69:
@soundanalogous How about now?

soundanalogous:
perfect. thank you!

PR #311, opened 2016-10-16
Only disable PWM if previously enabled

soundanalogous:
Only disable PWM when setting pinMode to OUTPUT if pin mode was previously set to PWM.

Needs to be thoroughly tested with a few Firmata clients to make sure this doesn't introduce any undesired side effects. That will determine whether or not this change should be marked as a bug fix or if it necessitates as a minor version update.


soundanalogous:
I'm not seeing any side affects on Breakout.js

rwaldron:
I actually forgot that I was using this branch until I saw this commit https://github.com/firmata/arduino/commit/c744d22017acf1b6f03b6723fdcd5ffd70e67718

PR #312, opened 2016-10-17
add Teensy 3.5 and 3.6 to Boards.h

soundanalogous:
This is a first pass. Has not been tested with actual hardware.

Need to verify that analog pin mapping is correct.

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -343,6 +343,36 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p)
 
 
+// Teensy 3.5 and 3.6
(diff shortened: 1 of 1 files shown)


soundanalogous:
@PaulStoffregen FYI

soundanalogous:
One potential discrepancy I found is that in pins_arduino.h, [A10 and A11 are mapped to D64 and D65](https://github.com/PaulStoffregen/cores/blob/master/teensy3/pins_arduino.h#L101-L102), but in core_pins.h it appears that [A10 and A11 are mapped to D40 and D41](https://github.com/PaulStoffregen/cores/blob/master/teensy3/core_pins.h#L1270-L1271).

Issue #215, opened 2015-08-29
create general purpose c/c++ version of Firmata.cpp


soundanalogous:
Currently, code within Firmata.cpp directly calls many Arduino-specific functions. Create a Firmata base class that is purely c/c++ that can be hosted outside of /firmata/arduino and used therefore easily used for any c-based Firmata client.

For backwards compatibility (since many existing Arduino sketches include "Firmata.h"), the purely c/c++ version would have to be named something like FirmataBase. On the Arduino side, Firmata would either wrap or extend FirmataBase to include Arduino-specific functionality. Likewise for c-based Firmata clients or even non-Arduino architectures.

zfields:
I'm putting together a prototype for this work item.

soundanalogous:
Done via https://github.com/firmata/arduino/pull/313.

PR #313, opened 2016-10-28
Split parsing functionality into new class

zfields:
@soundanalogous #215 (this is for a code review only)

The next steps would be to remove the functionality contained in this new code block from Firmata[.h|.cpp] and restitch it back together.

### DIFF
--- a/Firmata.cpp
+++ b/Firmata.cpp
@@ -24,6 +24,20 @@ extern "C" {
 }
 
 //******************************************************************************
+//* Static Members
+//******************************************************************************
+// make one instance for the user to use
+FirmataClass Firmata;
+
+void printVersion (void) {
+   Firmata.printVersion();
+}
+
+void printFirmwareVersion (void) {
+   Firmata.printFirmwareVersion();
+}
+
+//******************************************************************************
 //* Support Functions
 //******************************************************************************
 
@@ -66,7 +80,8 @@ FirmataClass::FirmataClass()
   firmwareVersionCount = 0;
   firmwareVersionVector = 0;
   blinkVersionDisabled = false;
-  systemReset();
+  parser.attach(REPORT_FIRMWARE, ::printFirmwareVersion);
+  parser.attach(REPORT_VERSION, ::printVersion);
 }
 
 //******************************************************************************
@@ -229,53 +244,13 @@ int FirmataClass::available(void)
 }
 
 /**
- * Process incoming sysex messages. Handles REPORT_FIRMWARE and STRING_DATA internally.
- * Calls callback function for STRING_DATA and all other sysex messages.
- * @private
- */
-void FirmataClass::processSysexMessage(void)
-{
-  switch (storedInputData[0]) { //first byte in buffer is command
-    case REPORT_FIRMWARE:
-      printFirmwareVersion();
-      break;
-    case STRING_DATA:
-      if (currentStringCallback) {
-        byte bufferLength = (sysexBytesRead - 1) / 2;
-        byte i = 1;
-        byte j = 0;
-        while (j < bufferLength) {
-          // The string length will only be at most half the size of the
-          // stored input buffer so we can decode the string within the buffer.
-          storedInputData[j] = storedInputData[i];
-          i++;
-          storedInputData[j] += (storedInputData[i] << 7);
-          i++;
-          j++;
-        }
-        // Make sure string is null terminated. This may be the case for data
-        // coming from client libraries in languages that don't null terminate
-        // strings.
-        if (storedInputData[j - 1] != '\0') {
-          storedInputData[j] = '\0';
-        }
-        (*currentStringCallback)((char *)&storedInputData[0]);
-      }
-      break;
-    default:
-      if (currentSysexCallback)
-        (*currentSysexCallback)(storedInputData[0], sysexBytesRead - 1, storedInputData + 1);
-  }
-}
-
-/**
  * Read a single int from the input stream. If the value is not = -1, pass it on to parse(byte)
  */
 void FirmataClass::processInput(void)
 {
   int inputData = FirmataStream->read(); // this is 'int' to handle -1 when no data
   if (inputData != -1) {
-    parse(inputData);
+    parser.parse(inputData);
   }
 }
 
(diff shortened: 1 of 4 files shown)


FirmataParser.h
```diff
@@ -0,0 +1,108 @@
...
+ * Query using the REPORT_VERSION message.
+ */
+#define FIRMATA_PROTOCOL_MAJOR_VERSION  2 // for non-compatible changes
+#define FIRMATA_PROTOCOL_MINOR_VERSION  5 // for backwards compatible changes
+#define FIRMATA_PROTOCOL_BUGFIX_VERSION 1 // for bugfix releases
+
+#define MAX_DATA_BYTES                  64 // max number of data bytes in incoming messages
+
+// Arduino 101 also defines SET_PIN_MODE as a macro in scss_registers.h
+#ifdef SET_PIN_MODE
```
soundanalogous: this should remain in Firmata.h

soundanalogous: I guess if it's move to Firmata.h it would need to be defined before FirmataParser is included. May be weird, but it is a hack in the first place.

zfields: It would be available to anyone who includes Firmata.h, via Firmata.h's inclusion of FirmataParser.h. Currently, it sits where SET_PIN_MODE is defined and used, and you wouldn't need to worry about any funky ordering.

zfields: Realistically, all your `#define`s should use the `#ifdef` `#undef` syntax for maximum portability of FirmataParser.h

FirmataParser.cpp
```diff
@@ -0,0 +1,300 @@
...
+    case DIGITAL_MESSAGE: currentDigitalCallback = newFunction; break;
+    case REPORT_ANALOG: currentReportAnalogCallback = newFunction; break;
+    case REPORT_DIGITAL: currentReportDigitalCallback = newFunction; break;
+    case SET_PIN_MODE: currentPinModeCallback = newFunction; break;
+    case SET_DIGITAL_PIN_VALUE: currentPinValueCallback = newFunction; break;
+  }
+}
+
+/**
+ * Attach a callback function for the SYSTEM_RESET command.
```
soundanalogous: this comment block needs to be updated since it now uses the more generic systemCallbackFunction

soundanalogous: Suggested text:
"Attach a system callback function (options are: REPORT_FIRMWARE, REPORT_VERSION and SYSTEM_RESET).
@param command The ID of the command to attach a callback function to.
@param newFunction A reference to the callback function to attach.

zfields: updating...

zfields:
@soundanalogous I have updated the issues you highlighted, and have updated FirmataClass to use FirmataParser. It passes all ArduinoUnit tests in the firmata_test test suite.

zfields:
I am using an UNO, and I have confirmed it compiles on all samples except:
- StandardFirmataBLE
- StandardFirmataChipKIT
- StandareFirmataWiFi

I think I'm missing external libraries.

zfields:
I ran a quick test using Windows Remote Arduino, and it is still able to parse the board description. However, I left my testing shield at work, and I won't be able to get it until tonight to confirm the analog capability. So I would leave this as DO NOT PULL, unless you have an end-to-end test you are comfortable with.

soundanalogous:
> I am using an UNO, and I have confirmed it compiles on all samples except:
> 
> StandardFirmataBLE
> StandardFirmataChipKIT
> StandareFirmataWiFi

Hummm... I would expect it to compile for StandardFirmataWiFi and perhaps also for StandardFirmataBLE. That is indicating a bug in those sketches. I'll look into it separately. Thanks for pointing that out.

Firmata.cpp
```diff
@@ -66,7 +80,8 @@ FirmataClass::FirmataClass()
   firmwareVersionCount = 0;
   firmwareVersionVector = 0;
   blinkVersionDisabled = false;
-  systemReset();
```
soundanalogous: Should parser.systemReset() be called here?

zfields: The constructor of FirmataParser should accomplish this same functionality, since there would not be any callback associated with FirmataClass at this time. I did consider this while I was working, but will you review the constructor of FirmataParser and confirm?

zfields: I should also mention the FirmataParser constructor will be implicitly called during the FirmataClass constructor.

soundanalogous: I see I had overlooked that. It appears that everything is properly initialized.

zfields:
@soundanalogous I've used my Firmata testing shield and everything looks good to me. Not only was I able to receive the `CAPABILITY_QUERY` (sysex), but I was also able to successfully receive both analog and digital values.
Also, the sending of digital and PWM signals worked, but aren't relevant to the parsing functionality.

At this point, I don't know what else to test. That being said, once we have completed your code review, I am ready for you to pull.

soundanalogous:
Would it be convenient to also break out the sysex constants to a separate file?

zfields:
My thought would be to keep this change as surgical as possible. Then once it's in place, it would provide the freedom to more easily make several other changes.

soundanalogous:
StandardFirmataBLE (nor any other variants) compile for the Arduino 101. I get the following error:

```
call of overloaded 'attach(int, void (&)(byte, int))' is ambiguous
```

The 101 uses a different compiler than the AVR and SAM based Arduino boards. You can replicate by using the Arduino Boards manager to install the Intel Curie Boards package and then compiling for the Arduino 101.

soundanalogous:
Actually Due, Zero, MKR1000, ESP8266, and the ChipKit boards also throw the same error. Seems any non-AVR compilers throw that error.

zfields:
Are those by chance, 32-bit boards? Ironically, I switched to `int16_t` instead of `int` on the callbacks, so 32-bit architectures would match the UNO execution exactly. I'll switch it back and I bet they will all compile; let me change it back.

zfields:
@soundanalogous I switched the type back, and now I'm able to get `StandardFirmataBLE.ino` to compile using the Arduino 101 compiler with no errors and no warnings.

soundanalogous:
There is one additional issue. The compiler for the Intel Galileo (v1 and v2) and Intel Edison boards does not understand `nullptr`. I've had so much trouble trying to maintain support for those 3 boards over the years since they've been available that I'm tempted to just drop Firmata support for them altogether.

zfields:
Although I have found Intel's Arduino "support" to be frustrating myself, I don't want to be the straw that breaks the camels back. I am near certain that `nullptr` equates to `(void *)NULL`. I'll put this change in and I think we all win without sacrificing any functionality.

soundanalogous:
SG. I've also tested on an Uno with a bunch of sensors and actuators hooked up and see no issues.

zfields:
@soundanalogous All instances of `nullptr` have been replaced with a `NULL` cast to the appropriate pointer type, and all my tests are still passing.

soundanalogous:
It's compiling for all supported boards and passes my smoke test. Merging...

soundanalogous:
Thanks @zfields for all of your work on this!

soundanalogous:
I could also break FirmataParser out into a separate repo and then just copy the .cpp and .h files into Firmata and ConfigurableFirmata whenever FirmataParser is updated. Probably easier for others to include FirmataParser in c++ client libraries that way.

zfields:
I would love to see FirmataParser in its own repository. However, I would also like to see you move the `#define`s into their own `.h` file as you previously suggested. That would mean you would need to either move the `#define`s into their own repo, or you would have to duplicate the code (obviously I am not suggesting duplication). For now, I would let it stew.

zfields:
@soundanalogous It was great working with you! Thanks for helping me test and iterate on the PR so quickly! Cheers!

Issue #322, opened 2016-11-07
older avr-gcc compiler doesn't handle array initialization in constructor list


soundanalogous:
I'm trying maintain Firmata back to Arduino v1.0.6. I noticed the compiler doesn't know how to handle array initialization in the FirmataParser constructor list. I get the following error:

```
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp: In constructor 'FirmataParser::FirmataParser()':
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp:32: error: expected `(' before '{' token
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp:32: error: expected `;' before '}' token
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp:32: warning: statement has no effect
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp: At global scope:
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp:32: error: mixing declarations and function-definitions is forbidden
/Applications/arduino-1.0.6/Arduino.app/Contents/Resources/Java/libraries/Firmata/FirmataParser.cpp:33: error: expected constructor, destructor, or type conversion before '(' token
```

Worth even trying to find a work around to maintain support for older Arduino versions?

soundanalogous:
@zfields please take a look at this.

soundanalogous:
Fails on `storedInputData{0},` in the following:

```c++
FirmataParser::FirmataParser()
:
    executeMultiByteCommand(0),
    multiByteChannel(0),
    storedInputData{0},
    waitForData(0),
    parsingSysex(false),
    sysexBytesRead(0),
    currentAnalogCallback((callbackFunction)NULL),
    currentDigitalCallback((callbackFunction)NULL),
    currentReportAnalogCallback((callbackFunction)NULL),
    currentReportDigitalCallback((callbackFunction)NULL),
    currentPinModeCallback((callbackFunction)NULL),
    currentPinValueCallback((callbackFunction)NULL),
    currentReportFirmwareCallback((systemCallbackFunction)NULL),
[...]

PR #323, opened 2016-11-07
Update FirmataParser to support older compilers

zfields:
Addresses issue #322 

Passes all tests I have available


soundanalogous:
LGTM. Compiles without issue in Arduino 1.0.6 and in Arduino 1.6.12.

PR #326, opened 2016-11-09
Moving FirmataConstants inclusion from header

zfields:
Addresses issue #324

Issue #324, opened 2016-11-07
Move inclusion of FirmataConstants.h in FirmataMarshaller from .h to .cpp


zfields:
It is unnecessary to place `FirmataConstants.h` in the headers. It can do its job, making more readable code, on the `.cpp` side. If an end user elects to use the constants (as opposed to making their own to avoid naming conflicts) they are enabled to do so by including `FirmataConstants.h`.

zfields:
FirmataParser has a dependency on MAX_DATA_BYTES, and therefore cannot have FirmataConstants.h removed from the `.h` file

zfields:
This is addressed in the most recent PR.

PR #327, opened 2016-11-10
Compiler optimizations for new classes

zfields:
PR provides compiler optimizations by labeling the functions as `const`.
Also replaces #326 by addressing issue #324 

Only optimizes the internal functionality, and does not affect the external API.

### DIFF
--- a/Firmata.h
+++ b/Firmata.h
@@ -111,7 +111,7 @@ class FirmataClass
 
     /* private methods ------------------------------ */
     void strobeBlinkPin(byte pin, int count, int onInterval, int offInterval);
-    friend void FirmataMarshaller::sendValueAsTwo7bitBytes(uint16_t value);
+    friend void FirmataMarshaller::sendValueAsTwo7bitBytes(uint16_t value) const;
 };
 
 extern FirmataClass Firmata;
--- a/FirmataMarshaller.cpp
+++ b/FirmataMarshaller.cpp
@@ -23,6 +23,8 @@
   #include <string.h>
 #endif
 
+#include "FirmataConstants.h"
+
 //******************************************************************************
 //* Support Functions
 //******************************************************************************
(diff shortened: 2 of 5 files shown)


zfields:
@soundanalogous Did you have any thoughts or concerns about this PR?

soundanalogous:
I haven't had enough time to look into it. Will look at it tomorrow. I generally do most of my work on Firmata on weekends and just check in from time to time during the week.

zfields:
My apologies for rushing you, I'm just excited and iterating quickly. I'll dial it down a notch. :smiley:

Firmata.h
```diff
@@ -111,7 +111,7 @@ class FirmataClass
 
     /* private methods ------------------------------ */
     void strobeBlinkPin(byte pin, int count, int onInterval, int offInterval);
-    friend void FirmataMarshaller::sendValueAsTwo7bitBytes(uint16_t value);
+    friend void FirmataMarshaller::sendValueAsTwo7bitBytes(uint16_t value) const;
```
soundanalogous: What is the compiler optimization of this use of `const`. I was not familiar with this technique and just looked into it. From what I understand it informs the compiler that no class members within the scope of this method can be altered. I'm just not sure what the advantage of this declaration is.

zfields: From the C++ Standard:

> **9.3.2  The this pointer**
> ...
> If the member function is declared `const`, the type of this is `const X*`, if the member
> function is declared `volatile`, the type of this is `volatile X*`, and if the member function
> is declared `const volatile`, the type of this is `const volatile X*`. *[Note: thus in a `const`
> member function, the object for which the function is called is accessed through a `const`
> access path. — end note]*

There is no guarantee, but the compiler may able to optimize around constants.

Here is an excellent [discussion on StackOverflow](http://stackoverflow.com/questions/27466642/what-kind-of-optimization-does-const-offer-in-c-c-if-any).

Here a well written [blog by Herb Sutter](http://www.gotw.ca/gotw/081.htm)

soundanalogous:
Thanks for the links. I've tested this and there are no errors so merging.

PR #336, opened 2016-12-06
enable baud rate for serial feature to be updated

painengineering:


utility/SerialFirmata.h
```diff
@@ -65,6 +65,7 @@
 #define SERIAL_CLOSE                0x50
 #define SERIAL_FLUSH                0x60
 #define SERIAL_LISTEN               0x70
+#define SERIAL_UPDATE_BAUD          0x80
```
soundanalogous: Data bytes cannot be > 0x7F in Firmata. I made the same mistake when I drafted FirmataSerial. I'm proposing a 2.0 release to make this feature more scalable. It will not be backwards compatible, but it's the best way to move forward here IMO: https://github.com/firmata/protocol/pull/71.

PR #337, opened 2016-12-14
Add support for Arduino MKRZero

sandeepmistry:
Adds support for the new [Arduino MKRZero](https://www.arduino.cc/en/Main/ArduinoBoardMKRZero) board.

This board has some gaps in the usable pins, the built-in LED is on pin 32 and the ADC for the battery is on pin 33. Pins 22 - 31 are used for USB, AREF, the SD card reader (SPI), and a pad on the bottom of the board.

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -260,6 +260,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
 
+// Arduino MKRZero
+#elif defined(ARDUINO_SAMD_MKRZERO)
+#define TOTAL_ANALOG_PINS       7
+#define TOTAL_PINS              34 // 8 digital + 3 spi + 2 i2c + 2 uart + 7 analog + 3 usb + 1 aref + 5 sd + 1 bottom pad + 1 led + 1 battery adc
+#define IS_PIN_DIGITAL(p)       ((((p) >= 0 && (p) <= 21) || (p) == 32) && !IS_PIN_SERIAL(p))
+#define IS_PIN_ANALOG(p)        (((p) >= 15 && (p) < 15 + TOTAL_ANALOG_PINS) || (p) == 33)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS) // deprecated since v2.4
+#define IS_PIN_I2C(p)           ((p) == 11 || (p) == 12) // SDA = 11, SCL = 12
+#define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
+#define IS_PIN_SERIAL(p)        ((p) == PIN_SERIAL1_RX || (p) == PIN_SERIAL1_TX) //defined in variant.h  RX = 13, TX = 14
+#define PIN_TO_DIGITAL(p)       (p)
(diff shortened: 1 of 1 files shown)


sandeepmistry:
I forgot to mention, this will work with the next SAMD core release. As the v1.6.9 release of the SAMD core incorrectly has the board define for the MKRZero set to MKR1000 (https://github.com/arduino/ArduinoCore-samd/commit/84ccd3118483f75eddd628c78f589457fa01c1b0).

cc/ @cmaglie

Boards.h
```diff
@@ -260,6 +260,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#elif defined(ARDUINO_SAMD_MKRZERO)
+#define TOTAL_ANALOG_PINS       7
+#define TOTAL_PINS              34 // 8 digital + 3 spi + 2 i2c + 2 uart + 7 analog + 3 usb + 1 aref + 5 sd + 1 bottom pad + 1 led + 1 battery adc
+#define IS_PIN_DIGITAL(p)       ((((p) >= 0 && (p) <= 21) || (p) == 32) && !IS_PIN_SERIAL(p))
+#define IS_PIN_ANALOG(p)        (((p) >= 15 && (p) < 15 + TOTAL_ANALOG_PINS) || (p) == 33)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS) // deprecated since v2.4
+#define IS_PIN_I2C(p)           ((p) == 11 || (p) == 12) // SDA = 11, SCL = 12
+#define IS_PIN_SPI(p)           ((p) == SS || (p) == MOSI || (p) == MISO || (p) == SCK)
+#define IS_PIN_SERIAL(p)        ((p) == PIN_SERIAL1_RX || (p) == PIN_SERIAL1_TX) //defined in variant.h  RX = 13, TX = 14
```
soundanalogous: Do PIN_SERIAL1_RX and _TX correlate to `Serial` or `Serial1`?

Boards.h
```diff
@@ -260,6 +260,23 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
 
+// Arduino MKRZero
+#elif defined(ARDUINO_SAMD_MKRZERO)
+#define TOTAL_ANALOG_PINS       7
+#define TOTAL_PINS              34 // 8 digital + 3 spi + 2 i2c + 2 uart + 7 analog + 3 usb + 1 aref + 5 sd + 1 bottom pad + 1 led + 1 battery adc
```
soundanalogous: Just checking that this number is the highest addressable pin number rather than the total number of usable pins. The name `TOTAL_PINS` is actually misleading for this reason.

sandeepmistry: [mkrzero/variant.cpp](https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrzero/variant.cpp#L125-L146) has entries for a 32768Hz crystal on pins 34 and 35, but I don't think Firmata users need to access them (MKR1000 also has those entries).

So the highest pin number that needs to be addressed is 33 for the Battery voltage ADC.

soundanalogous: sounds good. 34 is correct then since TOTAL_PINS would be last addressable pin number + 1 (because pin count starts at 0).

Boards.h
sandeepmistry: `Serial1` just like the MKR1000, `Serial` correlates to native USB.

soundanalogous:
Thanks for the PR. This is super helpful!

Issue #334, opened 2016-11-28
Multiple analog reads mess up ADC on Arduino UNO


webondevices:
There is a well-known problem with the Arduino UNO ADC (very likely affects other boards as well) that multiple analogRead() performed on different pins with no delay won't give enough time for the ADC to "cool down" and measurements will be completely wrong.

I discovered this issue while trying to make the LM35 temp sensor and an LDR measure correctly in the same circuit with the Johnny-Five library but they only work independently but not together.

In the Arduino community there are two solutions (hacks) to this:
- add a couple of ms delay between measurements on different pins
- perform two analogRead() on the same pin, dump the first and keep the second

I modified the StandardFirmata sketch by adding the dummy analogRead before the actual Firmata read and it solved all my problems:

```
dummyRead = analogRead(analogPin);
Firmata.sendAnalog(analogPin, analogRead(analogPin));
```

soundanalogous:
That's strange because the analogRead code in Arduino [blocks until the read has finished](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/wiring_analog.c#L81-L82) but I guess this doesn't account for an ADC channel change which can take a few microseconds.

The issue here is that the delay time is likely unique per microcontroller architecture (Firmata is supported on architectures other than AVR). I'm not sure how long a dummy read takes vs the minimum delay required to switch an ADC channel on a given microcontroller architecture. Until I have that data it's difficult to determine how to proceed.

soundanalogous:
It may actually be most effective to stagger reads, so instead of [reading all analog inputs in one shot](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/wiring_analog.c#L81-L82), stagger them, one read per iteration of the main loop, with the first read occurring at the sampling interval, assuming that all reads could occur within the sampling interval. This would not incur any additional delays. The only problem would be if the loop time is < the necessary channel change duration.

webondevices:
It's clear that the issue is not with the software implementation as the code blocks until the read has finished. This is most likely an issue with the ADC.

I wanted to highlight this thread on the Arduino forum where they are discussing this issue in a bit more detail:

http://forum.arduino.cc/index.php?topic=69675.0

webondevices:
[...]

PR #340, opened 2017-01-03
Added extra analogRead() to let ADC cool off

webondevices:
Proposed bug fix for: https://github.com/firmata/arduino/issues/334

**Summary of bug**
Reading multiple analog sensors the same time messes up all readings. For example: the LM35 works fine on its own reporting consistent values but as soon as another analog sensor is added it starts reporting random values between 12 - 50 C°. I tested this with multiple sensors with the same result.

The bug is most likely with the Arduino hardware as I managed to reproduce the issue with a simple two line Arduino sketch.

**Test**
I tested multiple analog sensors and combinations on a Genuine Arduino UNO and had the same issue all the time. I also tested on a Chinese Arduino NANO which had the same problem.

**My proposed solution have fixed issues on both Arduinos**

I only noticed at the end that apparently both Arduinos have an ATMEL MEGA 328P on board so it would be great to test the solution on other chips as well but I don't have access to any other boards at the moment.

### DIFF
--- a/examples/StandardFirmata/StandardFirmata.ino
+++ b/examples/StandardFirmata/StandardFirmata.ino
@@ -805,6 +805,8 @@ void loop()
       if (IS_PIN_ANALOG(pin) && Firmata.getPinMode(pin) == PIN_MODE_ANALOG) {
         analogPin = PIN_TO_ANALOG(pin);
         if (analogInputsToReport & (1 << analogPin)) {
+          // Extra analogRead() to cover ADC cooling-off period for higher accuracy
+          analogRead(analogPin);
           Firmata.sendAnalog(analogPin, analogRead(analogPin));
         }
       }


examples/StandardFirmata/StandardFirmata.ino
```diff
@@ -805,6 +805,8 @@ void loop()
       if (IS_PIN_ANALOG(pin) && Firmata.getPinMode(pin) == PIN_MODE_ANALOG) {
         analogPin = PIN_TO_ANALOG(pin);
         if (analogInputsToReport & (1 << analogPin)) {
+          // Extra analogRead() to cover ADC cooling-off period for higher accuracy
+          analogRead(analogPin);
```
soundanalogous: This is a waste of cycles if a user only has one analog input. However you could add a counter that increments each time you enter this block and then only if that count > 0 perform the extra analogRead:

```c++
if (analogReadCounter++ > 0) {
  analogRead(analogPin);
}
```

and reset analogReadCounter to 0 before the for loop.

webondevices:
Closing this due to new developments in the issue.

PR #341, opened 2017-01-08
Initial approach to a Firmata SPI implementation

soundanalogous:
This is an initial approach to Firmata SPI support. Consider this a prototype for now. The draft proposal that this implementation is based on is available here: https://github.com/firmata/protocol/blob/spi/spi-proposal.md. Also see comments in the proposal here for an overview on potential changes: https://github.com/firmata/protocol/pull/27/files.

SPI is a difficult protocol to implement in a generic way. The idea with this prototype implementation is to get it out there and have people start testing with various SPI devices to see where it falls short.

A prototype Firmata client implementation is also being explored for firmata.js: https://github.com/firmata/firmata.js/compare/spi-alpha. You can use that to test devices or use it as an example for testing in other programming environments.

Note that both the Arduino implementation and the firmata.js are still changing, but should be stable enough to checkout and test with (just be sure to checkout both at the same time as I'm pushing updates simultaneously).

### DIFF
--- a/utility/SPIFirmata.cpp
+++ b/utility/SPIFirmata.cpp
@@ -55,10 +55,10 @@ boolean SPIFirmata::handleSysex(byte command, byte argc, byte *argv)
       case SPI_BEGIN_TRANSACTION:
       {
         mDeviceId = argv[1] >> 2;
-        byte bitOrder = argv[2] & SPI_BIT_ORDER_MASK;
-        byte dataMode = argv[2] >> 1;
-        long clockSpeed = (long)argv[3] | ((long)argv[4] << 7) | ((long)argv[5] << 14) |
-                          ((long)argv[6] << 21) | ((long)argv[7] << 28);
+        uint8_t bitOrder = argv[2] & SPI_BIT_ORDER_MASK;
+        uint8_t dataMode = argv[2] >> 1;
+        uint32_t clockSpeed = (uint32_t)argv[3] | ((uint32_t)argv[4] << 7) | ((uint32_t)argv[5] << 14) |
+                          ((uint32_t)argv[6] << 21) | ((uint32_t)argv[7] << 28);
 
         if (argc > 8) {
           mCsPin = argv[8];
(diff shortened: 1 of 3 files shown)


utility/SPIFirmata.cpp
```diff
@@ -0,0 +1,211 @@
...
+void SPIFirmata::init()
+{
+  mDeviceId = 0;
+  mCsPin = -1;
+  mCsActiveState = SPI_CS_ACTIVE_LOW; // default
+}
+
+bool SPIFirmata::handlePinMode(uint8_t pin, int mode)
+{
+  // There is no reason for a user to manually set the SPI pin modes
```
rwaldron: Agreed.

utility/SPIFirmata.cpp
```diff
@@ -0,0 +1,211 @@
...
+            ((uint32_t)argv[5] << 14) | ((uint32_t)argv[6] << 21) | ((uint32_t)argv[7] << 28);
+
+        // argv[8] = wordSize, but not currently used since SPI.transfer only uses 8-bit words
+
+        if (argc > 9) {
+          mCsPin = argv[9];
+          pinMode(mCsPin, OUTPUT);
+
+          if (argv[10] != END_SYSEX) {
+            mCsActiveState = argv[10];
```
soundanalogous: Need to ensure that argv[10] is a valid value (0 or 1)

soundanalogous: fallback to default (SPI_CS_ACTIVE_LOW) if invalid

utility/SPIFirmata.cpp
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        SPISettings settings(clockSpeed, getBitOrder(bitOrder), getDataMode(dataMode));
+        SPI.beginTransaction(settings);
+        break;
+      }
+      case SPI_END_TRANSACTION:
+        SPI.endTransaction();
+        break;
+      case SPI_TRANSFER:
```
soundanalogous: TBD whether or not to keep this a single command or to break it up into 3 separate commands:

```
SPI_READ
SPI_WRITE
SPI_TRANSFER (normal SPI operation - read a word for every word written)
```

Another option would be to specify both an RX buffer size and a TX buffer size instead of specifying read/write modes. If only the RX buffer size is specified, then write 0x00 for each word to be read. If only the TX buffer size is specified, then write but ignore the return values. If TX buffer = RX buffer, replace the write word with the reply word and return the buffer. If RX buffer size is > TX buffer size, write TX words first, ignoring reply, then write 0x00 for each RX word. TBD how if there is a use case for RX buffer size < TX buffer size.

PR #344, opened 2017-01-12
Support analog (and motor) control on the ESP8266

ben--:
The ESP8266 doesn't play friendly with firmata or Johnny-Five because, in its default configuration, it expects PWM inputs to be in [0..1023] instead of [0..255].

The attached PR implements the simplest possible solution as a stopgap until a better solution can be found to resolve the general case for chips of varying resolutions.

### DIFF
--- a/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
+++ b/examples/StandardFirmataWiFi/StandardFirmataWiFi.ino
@@ -995,6 +995,9 @@ void initFirmata()
 void setup()
 {
   DEBUG_BEGIN(9600);
+#ifdef ESP8266
+  analogWriteRange(255);
+#endif
 
   initTransport();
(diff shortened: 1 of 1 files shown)


soundanalogous:
The issue is in firmata.js, not in the Arduino Firmata library. The Firmata Arduino library correctly reports the resolution, but firmata.js (which Johnny-Five uses to communicate with the Arduino board) does not yet report the resolution, but [there is a PR open](https://github.com/firmata/firmata.js/pull/144) to add this functionality.

rwaldron:
> in its default configuration, it expects PWM inputs to be in [0..1023] instead of [0..255].

To clarify: Firmata.js default resolution for analog _input_ is 10 bits (0-1023) and PWM _output_ is 8 bits (0-255).

Issue #253, opened 2015-12-19
Prepend all defines with "FIRMATA_"


soundanalogous:
New boards keep breaking Firmata (and likely other libraries) by including defines and macros with the same name in their libraries. Time to fix this once and for all and prepend "FIRMATA_" to all defines. This will break many sketches unfortunately (for example if people maintain their own version of StandardFirmata or something similar and upgrade to the version of the Firmata that changes the definitions). Probably best for 3.0 with incremental patches along the way.

soundanalogous:
Another approach would be to switch to namespaced enums for Firmata 3.0

DomAmato:
Namespaced enums is probably a safer route though for consistency prepending everything with FIRMATA_ would be easier and less confusing unless there was a strong naming convention for the namespacing

soundanalogous:
No longer necessary since we now have namespacing.

PR #348, opened 2017-02-24
namespace firmata

zfields:
Continues to expose the API that was previously exposed, but also allows a consumer to use the complete API fully encapsulated in the `firmata` namespace


zfields:
This allows firmata to pulled into bigger projects and ensures there will be no name collisions. This should be a no-op but it would still be good to test.

zfields:
I just confirmed it passes `firmata_test` on an Uno and it runs with the ESP8266 node sample.

zfields:
I initially did this for my project, but I just realized this addresses issue #253 

However this approach does not break your API, and therefore does not need to wait for v3.0. However, you are free to eliminate everything not in a namespace at v3.0 or as you see fit.

FirmataConstants.h
```diff
@@ -14,268 +14,84 @@
...
-#ifdef SYSEX_NON_REALTIME
-#undef SYSEX_NON_REALTIME
-#endif
-#define SYSEX_NON_REALTIME      0x7E // MIDI Reserved for non-realtime messages
-
-#ifdef SYSEX_REALTIME
-#undef SYSEX_REALTIME
-#endif
-#define SYSEX_REALTIME          0x7F // MIDI Reserved for realtime messages
+static const int SERIAL_MESSAGE =          0x60; // communicate with serial devices, including other boards
```
soundanalogous: do these each consume 16 bits?

Firmata.h
```diff
@@ -15,7 +15,7 @@
 #define Firmata_h
 
 #include "Boards.h"  /* Hardware Abstraction Layer + Wiring/Arduino */
-#include "FirmataConstants.h"
+#include "FirmataDefines.h"
```
soundanalogous: I don't think this filename change was intentional.

soundanalogous: oh nevermind... I see you split FimrataConstants into 2 files.

Firmata.h
```diff
@@ -92,10 +91,10 @@ class FirmataClass
     void write(byte c);
 
     /* attach & detach callback functions to messages */
-    void attach(uint8_t command, ::callbackFunction newFunction);
-    void attach(uint8_t command, ::systemCallbackFunction newFunction);
-    void attach(uint8_t command, ::stringCallbackFunction newFunction);
-    void attach(uint8_t command, ::sysexCallbackFunction newFunction);
+    void attach(uint8_t command, callbackFunction newFunction);
```
soundanalogous: when to use :: ?

soundanalogous:
Do all of these new declarations consume memory?

FirmataConstants.h
zfields: `#define`s are known as integer literals and treated as `int` so this shouldn't make any difference. Also both the `#define`s and `static const` will show up in the .data section of the memory map.

To confirm I ran side-by-side compiles looking at the "program storage space" and "dynamic memory". Neither change for the Uno, 101^ or ESP8266.
^101 only reports "program storage space"

zfields:
There should be no additional run-time tax for these declarations. They only amount to text, which is optimized away by the compiler.

FirmataConstants.h
zfields: Yes, but they always did. `#defines` in the context above are also known as integer constants. `static const int` and `#define 0x7F` require the same amount of space in the .data segment.

In a side-by-side compile, there is no change in the "program storage space" or "dynamic memory" usage for the Uno, 101^ and ESP8266.

*^The 101 only reports "program storage space".*

Firmata.h
zfields: `::` points to the global namespace. I tucked them into `FirmataClass` as they are related to its implementation and names like `systemCallbackFunction` can be ambiguous.

I left the globals in place so they can be referenced by existing sketches, but can be easily removed in the future (v3.0).

Firmata.h
zfields: This split allows big project users to have all the constants under the firmata namespace, while Arduino users get to keep the global preprocessor defines.

The global preprocessor defines are based off the constants, so there is a single point of reference and the defines can be removed at a later date (v3.0) if desired.

FirmataConstants.h
```diff
@@ -14,268 +14,84 @@
...
 
+namespace firmata {
 /* Version numbers for the Firmata library.
  * The firmware version will not always equal the protocol version going forward.
  * Query using the REPORT_FIRMWARE message.
  */
-#define FIRMATA_FIRMWARE_MAJOR_VERSION  2
-#define FIRMATA_FIRMWARE_MINOR_VERSION  5
-#define FIRMATA_FIRMWARE_BUGFIX_VERSION 4
+static const int FIRMATA_FIRMWARE_MAJOR_VERSION =  2;
```
soundanalogous: remove FIRMATA_ from this since it's no longer necessary due to the namespace (we'll need to keep it in the define however)

FirmataConstants.h
```diff
@@ -14,268 +14,84 @@
...
+static const int FIRMATA_FIRMWARE_BUGFIX_VERSION = 4;
 
 /* Version numbers for the protocol.  The protocol is still changing, so these
  * version numbers are important.
  * Query using the REPORT_VERSION message.
  */
-#define FIRMATA_PROTOCOL_MAJOR_VERSION  2 // for non-compatible changes
-#define FIRMATA_PROTOCOL_MINOR_VERSION  5 // for backwards compatible changes
-#define FIRMATA_PROTOCOL_BUGFIX_VERSION 1 // for bugfix releases
+static const int FIRMATA_PROTOCOL_MAJOR_VERSION =  2; // for non-compatible changes
```
soundanalogous: remove FIRMATA_

FirmataConstants.h
soundanalogous: This would also be a good time to align on terminology. I've been debating between _MESSAGE and _DATA and unifying. Traditionally it's been _DATA, but when I added Serial I thought _MESSAGE may make more sense since that's a midi term. I need to think about this some more.

FirmataConstants.h
zfields: complete

FirmataConstants.h
zfields: complete

FirmataConstants.h
zfields: As soon as you figure out how you want this handled, I will be happy to switch it up. Also happy to have this in place and be able to code against it!

IMO, everything adhering to a protocol is a message, and when you have bytes outside the protocol they are data. I think of data as bytes that pass through a transport (i.e. Serial). Don't know if that provides any clarity, but thats what you made me think of.

Firmata.h
zfields: Also these names can be changed at any time before you stamp it with a version.

FirmataConstants.h
zfields: Also, if you need more time to think, these names can be changed at any time before you stamp it with a version.

zfields: @soundanalogous After reading the MIDI spec this morning, I'm going to take another stab at this. I think of the framing bytes and its payload as a message. Although the data encapsulated by the Firmata protocol is indeed a serial message, the actual description of the payload itself is serial data. That being said, my vote is for `SERIAL_DATA`.

soundanalogous: SGTM and keeps requires far fewer changes moving forward. I'll update in the protocol documentation as well.

zfields: I've updated `SERIAL_MESSAGE` to `SERIAL_DATA`. I think `DIGITAL_MESSAGE` and `ANALOG_MESSAGE` are named correctly. Are there any others we should change while we are doing this?

soundanalogous: that's all I can think of for now

soundanalogous:
Ran this through all of my tests with no issues. @zfields good to go unless there is anything else you wanted to add here.

zfields:
@soundanalogous I am ready for you to pull. The only thing not secured by the `firmata` namespace are the contents of Boards.h. However Boards.h is a very Firmata server (Arduino-side) specific file and provides no value (as I can see) to anyone wishing to use this a Firmata client library.

Issue #349, opened 2017-02-24
FirmataMarshaller does not contain CORE functionality


zfields:
Previously, we discussed that `FirmataMarshaller` and `FirmataParser` should only contain the core functionality and nothing more. Then in the future (think configurable Firmata) we would use the marshaller and parser as a base to extend.

Would you enumerate the core functionality missing from the marshaller and parser here, so I may use it as a checklist to implement the remaining methods?

zfields:
As an example...
`sendPinStateQuery` is missing from `FirmataMarshaller`

zfields:
@soundanalogous I would love your input on this list.

soundanalogous:
I'll work on the list sometime this weekend. However one thing this makes me think about is now that we have FirmataMarshaller if it makes sense to handle any Sysex commands in FirmataParser anymore. For example, could we move `STRING_DATA` and `REPORT_FIRMWARE` out of FirmataParser and handle them in FirmataMarshaller? That would keep FirmataParser more in line with midi and FirmataMarshaller handles the second layer protocol which is everything composed within sysex messages.

soundanalogous:
The following core functionality is not yet implemented in FirmataMarshaller:

- [ ] sendPinStateQuery
- [ ] sendExtendedAnalog (could alternatively combine within sendAnalog and then only add a private `sendExtendedAnalog` method)
- [ ] setSamplingInterval

zfields:
I think it would be good to move STRING_DATA and REPORT_FIRMWARE out of FirmataParser and into FirmataClass. This would allow for the elimination of the callbacks that belong to these.

However, I don't think they have a home in FirmataMarshaller.

soundanalogous:
I guess one advantage of keeping STRING_DATA and REPORT_FIRMWARE out of FirmataClass is they don't depend on any particular hardware, so maybe it's best to keep them in FirmataParser for now then.

zfields:
I think it depends on if you are planning to make callbacks for all known messages (especially core functionality), or if you want to allow the user to implement their own callback system through the sysex callback.

Right now in my project, I am using the callbacks that are available, but when they are not defined (i.e. capQuery, analogMapping, etc.) I'm not really having any problem capturing my sysex messages and parsing them myself using the constants in FirmataConstants.h.

[...]

PR #350, opened 2017-02-24
Complete base API

zfields:
This commit is intended to be the first of many addressing any missing APIs listed in issue #349

### DIFF
--- a/FirmataMarshaller.cpp
+++ b/FirmataMarshaller.cpp
@@ -266,7 +266,7 @@ const
 void FirmataMarshaller::sendPinStateQuery(uint8_t pin)
 const
 {
-  sendSysex(PIN_STATE_QUERY, 1, &pin);
+  sendSysex(PIN_STATE_QUERY, sizeof(pin), &pin);
 }
 
 /**
@@ -297,4 +297,15 @@ void FirmataMarshaller::sendString(const char *string)
 const
 {
   sendSysex(STRING_DATA, strlen(string), (uint8_t *)string);
+}
+
+/**
+ * The sampling interval sets how often analog data and i2c data is reported to the client.
+ * @param interval_ms The interval (in milliseconds) at which to sample
+ * @note The default sampling interval is 19ms
+ */
+void FirmataMarshaller::setSamplingInterval(uint16_t interval_ms)
+const
+{
+  sendSysex(SAMPLING_INTERVAL, sizeof(interval_ms), &interval_ms);
 }
--- a/FirmataMarshaller.cpp
+++ b/FirmataMarshaller.cpp
@@ -160,6 +160,28 @@ void FirmataMarshaller::end(void)
 //******************************************************************************
 
 /**
+ * Query the target's firmware name and version
+ */
+void FirmataMarshaller::queryFirmwareVersion(void)
+const
+{
+  if ( (Stream *)NULL == FirmataStream ) { return; }
+  FirmataStream->write(START_SYSEX);
+  FirmataStream->write(REPORT_FIRMWARE);
+  FirmataStream->write(END_SYSEX);
+}
+
+/**
+ * Query the target's Firmata protocol version
+ */
+void FirmataMarshaller::queryVersion(void)
+const
+{
+  if ( (Stream *)NULL == FirmataStream ) { return; }
+  FirmataStream->write(REPORT_VERSION);
+}
+
+/**
  * Halt the stream of analog readings from the Firmata host application. The range of pins is
  * limited to [0..15] when using the REPORT_ANALOG. The maximum result of the REPORT_ANALOG is limited to 14 bits
  * (16384). To increase the pin range or value, see the documentation for the EXTENDED_ANALOG
@@ -224,7 +246,6 @@ void FirmataMarshaller::sendAnalog(uint8_t pin, uint16_t value)
 const
 {
   if ( (Stream *)NULL == FirmataStream ) { return; }
-
   if ( (0xF >= pin) && (0x3FFF >= value) ) {
     FirmataStream->write(ANALOG_MESSAGE|pin);
     transformByteStreamToMessageBytes(sizeof(value), reinterpret_cast<uint8_t *>(&value), sizeof(value));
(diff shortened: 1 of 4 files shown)


FirmataMarshaller.cpp
```diff
@@ -255,6 +255,19 @@ const
...
+ * Send a pin state query to the Firmata host application. The resulting sysex message will have
+ * a PIN_STATE_RESPONSE command byte, followed by the pin number, the pin mode and a stream of
+ * bits to indicate any *data* written to the pin (pin state).
+ * @param pin The pin to query
+ * @note The pin state is any data written to the pin (i.e. pin state != pin value)
+ */
+void FirmataMarshaller::sendPinStateQuery(uint8_t pin)
+const
+{
+  sendSysex(PIN_STATE_QUERY, sizeof(pin), &pin);
```
soundanalogous: `sendSysex` will send `pin` as 2 bytes.

zfields: great catch! fixed.

FirmataMarshaller.cpp
```diff
@@ -66,14 +66,57 @@ const
...
+  FirmataStream->write(END_SYSEX);
+}
+
+/**
+ * Transform 8-bit stream into 7-bit message
+ * @param bytec The number of data bytes in the message.
+ * @param bytev A pointer to the array of data bytes to send in the message.
+ * @param max_bytes Force message to be n bytes, regardless of data bits.
+ */
+void FirmataMarshaller::transformByteStreamToMessageBytes (size_t bytec, uint8_t * bytev, size_t max_bytes)
```
soundanalogous: There is a similar Encoder7Bit class in ConfigurableFirmata that both reads and writes: https://github.com/firmata/ConfigurableFirmata/blob/master/src/Encoder7Bit.cpp

soundanalogous: and example use to [read](https://github.com/firmata/ConfigurableFirmata/blob/master/src/OneWireFirmata.cpp#L108) and [write](https://github.com/firmata/ConfigurableFirmata/blob/master/src/OneWireFirmata.cpp#L68-L75)

zfields: Cool, I wish I had seen it before! ;-) However, just looking at it (and mulling it over) briefly, it looks like another dependency that will need to be managed.

Since this is a self-contained, private function, we will be able to refactor it at any time, without impacting the external API. For now, I would like to move forward using this function, until ConfigurableFirmata becomes a little more mainstream (or Encoder7Bit is refactored to consume/extend the FirmataMarshaller API) and gets pulled under the `firmata` namespace.

zfields:
@soundanalogous I responded inline to your suggestion. In short, I think it's a great idea, but is non-blocking and would be better suited during a larger refactoring.

Otherwise, I have tested this with every means I have available, and everything looks good.
- Windows Remote Arduino -> Uno
- firmata_tests.ino
- ESP8266 StandardFirmataWiFi from node.

FirmataMarshaller.cpp
```diff
@@ -173,17 +216,20 @@ const
  * when using the ANALOG_MESSAGE. The maximum value of the ANALOG_MESSAGE is limited to 14 bits
  * (16384). To increase the pin range or value, see the documentation for the EXTENDED_ANALOG
  * message.
- * @param pin The analog pin to send the value of (limited to pins 0 - 15).
+ * @param pin The analog pin to which the value is sent.
  * @param value The value of the analog pin (0 - 1024 for 10-bit analog, 0 - 4096 for 12-bit, etc).
- * The maximum value is 14-bits (16384).
```
soundanalogous: Add this back in with the updated max. I assume it's 32 bits. However different firmata client libraries may interpret this in different ways (I think some only expect a max of 16 bits), but that shouldn't break anything unless a library looks for a specific number of bytes rather than accepting all bytes up to END_SYSEX (but that would be the fault of the client library).

FirmataMarshaller.cpp
```diff
@@ -173,17 +216,20 @@ const
...
  */
 void FirmataMarshaller::sendAnalog(uint8_t pin, uint16_t value)
 const
 {
   if ( (Stream *)NULL == FirmataStream ) { return; }
-  // pin can only be 0-15, so chop higher bits
-  FirmataStream->write(ANALOG_MESSAGE | (pin & 0xF));
-  sendValueAsTwo7bitBytes(value);
+
+  if ( (0xF >= pin) && (0x3FFF >= value) ) {
```
soundanalogous: this should be `||` rather than `&&` since it's possible to have a pin number > 15 with a value > 14 bits

soundanalogous: oh wait, this is fine as is. I'm not used to the static value being on the left side of the comparison so it confused me for a second.

zfields: Ha! Yeah, I used to be teaching assistant, so I always taught the kids to put it on the left because you can't accidentally set it when trying to compare.

FirmataMarshaller.cpp
zfields: This did not change. I removed it when I thought it would be changing. I have replaced the note.

zfields:
Should we add something around the reset functionality while we are doing this? Got any names in mind?

FirmataMarshaller.cpp
soundanalogous: It's limited to 14 bits when using ANALOG_MESSAGE, but sendAnalog now calls sendExtendedAnalog so larger values should be allowed. No cap is defined in the protocol documentation but I think 32-bit is probably reasonable (although not sure I've ever seen hardware that supports that resolution).

soundanalogous: here's the equivalent in firmata.js for example: https://github.com/firmata/firmata.js/blob/master/lib/firmata.js#L776-L806

soundanalogous:
It's purely a software reset. A user crafting their own Firmata sketch can do anything they want in the callback. StandardFirmata initializes everything to a known state. FirmataParser resets the buffer.

So naming is a bit tricky. `systemReset` could imply a HW reset so that may be misleading. Perhaps `softReset` or maybe `systemReset` is okay and we just document that it doesn't reset the physical board, it simply resets the buffer and the user can do anything else they want in the systemResetCallback.

zfields:
Please review the three new functions `queryFirmwareVersion`, `queryVersion`, `systemReset`.
I have rerun all the tests I am able to and I'm still passing. I think we should pull this in (as the marshalling functionality is "complete"), and decide the fate of the parser (whether to add or subtract) over the next week/weekend.

soundanalogous:
Sounds good. Have you compiled against a bunch of boards in the IDE (even if you don't have those actual boards - including ones like the Intel boards which have the fussier c++ compiler)?

zfields:
It compiles on ESP8266, 101 and AVR without warning (at least not caused by me ;-) ).

zfields:
Just downloaded the SAMD toolchain, and it builds for MKR1000

soundanalogous:
Ran through my usual compile tests in Arduino 1.8.1 (I really need to spend time to figure out a way to automate this at some point). Also tested with Arduino 1.6.12, 1.5.8 and 1.0.6. Fails with Arduino 1.6.11 but it appears to be an issue with arduino-builder, not Firmata (have noticed this previously as well... I guess Arduino is not testing older versions anymore, or maybe they only test the most recent 2 versions).

One thing to be mindful of is the increase in program memory. Compiling StandardFirmataPlus for an Arduino Uno in Arduino 1.8.1 with Firmata v2.5.4 consumes 13924 bytes of Flash and 1244 bytes of SRAM. However with this most recent PR we are up to 15278 bytes of Flash (increase of 1354 bytes - this is all changes since v2.5.4 was released... not just this PR) and 1298 bytes of RAM. I'm not too concerned about the increased RAM since it's only +54 bytes, but we should keep and eye on the Flash increase since we've added 1354 bytes so far without adding any new Firmata features.

zfields:
Fair point. I expect much of this has to do with allowing the callbacks to have context, which automatically doubles the storage required, but allows for multiple instances and provides enormous functional gains. Depending on where you are looking to go with the library I could see it taking on a little more water or shedding a little weight. However, I believe a true refactor can produce significant memory footprint improvement. It's usually stuff you don't expect, but cutting functionality is usually not necessary.

soundanalogous:
Yeah I figure we can reduce usage through a refactor in the future. Cleaning up the StandardFirmata variants at some point will probably also help bring the Flash usage size down.

Issue #353, opened 2017-03-12
[v3.0.0] Sysex String Callback


zfields:
https://github.com/firmata/arduino/blob/master/FirmataParser.cpp#L420

This seems like it should be passing the string without the from index 1 instead of 0. It is a small and easy change, but it would change the behavior*, so it probably needs to wait until v3.

**It has always been this way...*

soundanalogous:
It shouldn't make any difference to change since the first byte is skipped here: https://github.com/firmata/arduino/blob/master/FirmataParser.cpp#L403. Couldn't you simply change i to 0 on line 403 and then change the index to 1 on line 420? Either way I think achieves the same result.

soundanalogous:
I wonder if there was a regression when you refactored to use `bufferDataAtPosition`. The way I used to do at (and in hindsight I have no idea why I handled it this way) was shift the data back one byte in the `storedInputData array`: https://github.com/firmata/arduino/blob/2.5.3/Firmata.cpp#L246-L254

zfields:
If it is a regression then we should just change it back (#356).

soundanalogous:
I'll have to test it to know for sure. The string callback isn't used in any of the StandardFirmata variants, so I overlooked it during all of the refactoring work.

soundanalogous:
I guess a unit test would catch this as well.

zfields:
We should definitely have unit tests, they would have shown us that it is correct! :-P

PR #356, opened 2017-03-12
Send only string in sysex string callback

zfields:


zfields:
I have only observed this behavior in code, I have not actually tested it to confirm. Please do not pull without checking first.

zfields:
To address #353

zfields:
tricky... finkle is einhorn -> einhorn is finkle, where i = einhorn and j = finkle

look close, the first think it does it move the byte at index 1 to 0.

PR #358, opened 2017-03-12
version callbacks

zfields:


zfields:
This causes some conflict with the node sample on the ESP8266. I don't know node all that well, so I'm not confident that I can debug it.

soundanalogous:
firmata.js doesn't use FirmataParser at all. It simply parses based on the way Firmata.cpp used to do things.

soundanalogous:
I have to admit I'm starting to get lost in all of these callbacks.

Does this codepath still run in some form? https://github.com/firmata/arduino/blob/2.5.4/Firmata.cpp#L240.

And this?
https://github.com/firmata/arduino/blob/2.5.4/Firmata.cpp#L369

FirmataMarshaller.cpp
```diff
@@ -168,6 +168,8 @@ const
   if ( (Stream *)NULL == FirmataStream ) { return; }
   FirmataStream->write(START_SYSEX);
   FirmataStream->write(REPORT_FIRMWARE);
+  FirmataStream->write(firmata::FIRMWARE_MAJOR_VERSION);
```
soundanalogous: This is a query for the firmware version so you don't know the version at this point.

FirmataMarshaller.cpp
```diff
@@ -179,6 +181,8 @@ const
 {
   if ( (Stream *)NULL == FirmataStream ) { return; }
   FirmataStream->write(REPORT_VERSION);
+  FirmataStream->write(firmata::PROTOCOL_MAJOR_VERSION);
```
soundanalogous: same here

FirmataMarshaller.cpp
soundanalogous: Maybe what is needed here instead is two new Firmata commands `QUERY_FIRMWARE_VERSION` and `QUERY_PROTOCOL_VERSION` and those have no payload. Only the Reply has data.

zfields: I totally agree. Currently, these additional bytes would be discarded by FirmataClass, however they cause the callbacks to be invoked correctly.

FirmataMarshaller.cpp
zfields: Yup, just fluff.

zfields:
To answer your previous question. Yes those code paths get executed, but in the static callbacks of FirmataClass.h.

https://github.com/firmata/arduino/blob/master/Firmata.h#L153
https://github.com/firmata/arduino/blob/master/Firmata.h#L154

FirmataMarshaller.cpp
soundanalogous: I still don't understand. These bytes aren't specified in the protocol for the query.

soundanalogous: I guess it doesn't matter since the firmware doesn't even need this. Only c++ client libraries.

FirmataParser.cpp
```diff
@@ -130,6 +130,9 @@ void FirmataParser::parse(uint8_t inputData)
           if (currentReportDigitalCallback)
             (*currentReportDigitalCallback)(currentReportDigitalCallbackContext, multiByteChannel, dataBuffer[0]);
           break;
+        case REPORT_VERSION:
+          if (currentReportVersionCallback)
+            (*currentReportVersionCallback)(currentReportVersionCallbackContext, dataBuffer[0], dataBuffer[1], (const char *)NULL);
```
soundanalogous: Firmata client libraries do not provide data parameters so I'm not sure where the version data would come from. You could pass it in statically though.

FirmataParser.cpp
```diff
@@ -394,14 +413,24 @@ void FirmataParser::processSysexMessage(void)
 {
   switch (dataBuffer[0]) { //first byte in buffer is command
     case REPORT_FIRMWARE:
-      if (currentReportFirmwareCallback)
-        (*currentReportFirmwareCallback)(currentReportFirmwareCallbackContext);
+      if (currentReportFirmwareCallback) {
+        size_t sv_major = dataBuffer[1], sv_minor = dataBuffer[2];
```
soundanalogous: same here. This data would never be in the buffer with existing Firmata client libraries.

FirmataMarshaller.cpp
zfields: Yes, that's right. However, if someone would like to get the bytes via the parser (theoretically, this would be the primary way), then we need to change the parser to be able to consume the additional bytes that provide the version information and change the protocol to provide them (which it currently does not).

As you pointed out, this could be accomplished by creating `QUERY_FIRMWARE_VERSION` and `QUERY_PROTOCOL_VERSION`, then migrating logic for `REPORT_FIRMWARE` and `REPORT_VERSION` over to them and keep the changes I've proposed to handle the actual reporting of versions and not the query.

FirmataParser.cpp
soundanalogous: Couldn't you just use the constants: https://github.com/firmata/arduino/blob/master/FirmataConstants.h#L22-L23?

zfields: No. I would have compiled in the constants, they describe the version on my side of the protocol not the other side.

FirmataParser.cpp
zfields: I am aware, but I need to standardize the message so the parser can parse it - even if portions are disregarded. The name of the byte is REPORT version

soundanalogous: I think I'm starting to understand. Let me know if this is correct.

The version parameters are needed in the callback since the callback is actually for the reply rather than the query. When the callback is called, those parameters are available.

The confusion on the protocol side is that `REPORT_VERSION` and `REPORT_FIRMWARE` where shared for both query and reply (which was common in the early days of Firmata).

zfields: Yes!

soundanalogous: I see since we can anticipate that dataBuffer[0] and dataBuffer[1] will contain the appropriate data when the callback is called. I was overlooking the fact that dataBuffer was a member variable, that's what was largely throwing me off. Maybe we should have a convention member variables. I've been prefixing with `m` for the SPI feature I've been working on: https://github.com/firmata/arduino/blob/spi-alpha/utility/SPIFirmata.h#L94-L96.

zfields: Originally, the parser was built into FirmataClass which serves the host/server.
There was no notion of a callback for version, ever; it was tightly coupled.
`REPORT_VERSION` was used to query the version from the host/server, and it would subsequently report the version using `REPORT_VERSION`.
The fact that `REPORT_VERSION` is a single byte operation in the protocol documentation is wrong, because when the version is actually reported it sends three bytes.
Therefore, `REPORT_VERSION` should always send three bytes and the server/host should be smart enough realize it is being queried and discard them.
An alternate approach would be to create a `QUERY_VERSION` single byte used to explicitly query the version from the server/host

soundanalogous: I'll add `QUERY_FIRMWARE_VERSION` and `QUERY_PROTOCOL_VERSION` for v3.0 since I can't really think of a good reason for a client library to send it's version to the board.

zfields: I prefix all my member variables, so I'm down. I was just following style.

FirmataParser.cpp
```diff
@@ -163,8 +166,8 @@ void FirmataParser::parse(uint8_t inputData)
         systemReset();
         break;
       case REPORT_VERSION:
-        if (currentReportVersionCallback)
-          (*currentReportVersionCallback)(currentReportVersionCallbackContext);
+        waitForData = 2; // two data bytes needed
```
soundanalogous: This is where I wish we had unit test coverage for the parser. I'm still not 100% sure this will not break with most existing Firmata client libraries.

soundanalogous: I'd want to know that if waitForData is 2 or 0 this will still work as expected.

soundanalogous: Yeah tracing through the code, this will cause the next two bytes after a `REPORT_VERSION` command is sent (as a query) from a client library to be skipped. So if `REPORT_VERSION` is followed immediately by a Sysex message (which I believe is the case in firmata.js) then that entire Sysex message will be lost (since the START_SYSEX and COMMAND bytes will be skipped).

soundanalogous: Actually `START_SYSEX` will get through because `( (waitForData > 0) && (inputData < 128) )`
But I don't think the COMMAND byte will because `waitForData` would still equal 2. One work around is in the START_SYSEX case, reset waitForData to 0.

zfields: It does change the behavior. With this change, you will have to send all three bytes for the version in order to get the callback from FirmataClass to fire.

soundanalogous: That's going to break a lot of client libraries. We'll have to wait for Firmata v3.0 then.

soundanalogous: This is because many Firmata client libraries initiate communication with the board by querying the protocol or version.

soundanalogous: I think there is some general confusion here because `REPORT_VERSION` is misleading. Basically a client should never report its version to the host/firmware. A client can only query the firmware and protocol versions from the host/firmware.

soundanalogous: But you're trying to use the same parser both on the host and client side so now I understand the larger issue here.

zfields:
I have rebased this branch on top of the recent `FIRMWARE_VERSION` update. Thus, resolving any obvious, or known, merge conflicts and making it more "cherry-pick"-able when v3.0.0 comes around.

PR #360, opened 2017-03-14
Added support for Arduino Primo board

chiararuggeri:
Hi,
I'm the software developer of the Arduino Primo board. I would like to add the support for this board in your library.
Thanks.

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -299,6 +299,21 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
+// Arduino Primo
+#elif defined(ARDUINO_PRIMO)
+#define TOTAL_ANALOG_PINS       6
+#define TOTAL_PINS              22 //14 digital + 6 analog + 2 i2c
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < 20)
+#define IS_PIN_ANALOG(p)        ((p) >= 14 && (p) < 20)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS+2)
+#define IS_PIN_I2C(p)           ((p) == PIN_WIRE_SDA || (p) == PIN_WIRE_SCL) // SDA = 20, SCL = 21
+#define IS_PIN_SPI(p)           ((p) == SS || (p)== MOSI || (p) == MISO || (p == SCK)) // 10, 11, 12, 13
+#define PIN_TO_DIGITAL(p)       (p)
+#define PIN_TO_ANALOG(p)        ((p) - 14)
+#define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
+#define PIN_TO_SERVO(p)         (p)
 
 // Arduino 101
 #elif defined(_VARIANT_ARDUINO_101_X_)


soundanalogous:
Do you have a link to the pins_arduino.h file and other pin definitions for the Primo?

Boards.h
```diff
@@ -299,6 +299,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
+// Arduino Primo
+#elif defined(NRF52)
```
soundanalogous: Use a more specific name than NRF52, otherwise this block may be defined for other NRF52-based boards in the future. I assume there is a define like "ARDUINO_<ARCH>_PRIMO" for this board.

Boards.h
```diff
@@ -299,6 +299,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define TOTAL_ANALOG_PINS       6
+#define TOTAL_PINS				22 //14 digital + 6 analog + 2 i2c
```
soundanalogous: fix the indent here, and for IS_PIN_SPI and IS_PIN_SERIAL

soundanalogous:
This will simply add Firmata support for the Primo over a serial transport. A good next step would be to add it to StandardFirmataWiFi and StandardFirmataBLE (separate PRs for each).

soundanalogous:
I found the variant.h file.

Boards.h
```diff
@@ -299,6 +299,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < 20)
+#define IS_PIN_ANALOG(p)        ((p) >= 14 && (p) < 20)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS+2)
+#define IS_PIN_I2C(p)           ((p) == 20 || (p) == 21)
+#define IS_PIN_SPI(p)			((p)== MOSI || (p) == MISO || (p == SCK)) // 11, 12, 13
+#define IS_PIN_SERIAL(p)		((p) == 0 || (p) == 1)
```
soundanalogous: Do pins 0 and 1 map to Serial or Serial1?

Boards.h
```diff
@@ -299,6 +299,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#elif defined(NRF52)
+#define TOTAL_ANALOG_PINS       6
+#define TOTAL_PINS				22 //14 digital + 6 analog + 2 i2c
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < 20)
+#define IS_PIN_ANALOG(p)        ((p) >= 14 && (p) < 20)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS+2)
+#define IS_PIN_I2C(p)           ((p) == 20 || (p) == 21)
+#define IS_PIN_SPI(p)			((p)== MOSI || (p) == MISO || (p == SCK)) // 11, 12, 13
```
soundanalogous: Based on variant.h, SPI pins are 27, 24, 26 respectively rather than 11, 12, 13

chiararuggeri:
I've added some of the fixes you requested. Regarding SPI pins, they are both 11, 12, 13 and 24, 26, 27. They are the same pins, enumerated twice (one for pin header and one for ICSP connector). Anyway if you prefer I will change them.
In Arduino Primo there is just one Serial connected to 0 and 1 pins. There is no Serial1 port.
Let me know if other modifications are needed.
Thanks

Boards.h
```diff
@@ -299,6 +299,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define TOTAL_ANALOG_PINS       6
+#define TOTAL_PINS              22 //14 digital + 6 analog + 2 i2c
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < 20)
+#define IS_PIN_ANALOG(p)        ((p) >= 14 && (p) < 20)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS+2)
+#define IS_PIN_I2C(p)           ((p) == PIN_WIRE_SDA || (p) == PIN_WIRE_SCL) // SDA = 20, SCL = 21
+#define IS_PIN_SPI(p)           ((p)== MOSI || (p) == MISO || (p == SCK)) // 11, 12, 13
+#define IS_PIN_SERIAL(p)        ((p) == 0 || (p) == 1)
```
soundanalogous: Remove the `IS_PIN_SERIAL` macro. This is for the Firmata Serial feature, but it's currently only support for Serial1, Serial2 and Serial3.

soundanalogous:
You can keep the SPI pins documented as 11, 12 and 13. This is helpful actually since that was not clear by looking at the variant.h file.

soundanalogous:
Also add a comment on the SPI line that SS = 10

chiararuggeri:
I've added the SS pin definition that I've forgotten. Missing something else?

soundanalogous:
That's it. If you comfortable squashing commits it would be nice to have those 6 commits squashed to a single commit. However if you're not familiar with that process, don't worry about it as it can be tricky.

chiararuggeri:
Squashed. Thank you for the support.

soundanalogous:
Thank you for your contribution.

Issue #363, opened 2017-03-18
Firmata strings encoded incorrectly


zfields:
This needs a patch rev [v2.5.6]. I encoded Firmata strings directly (not as two bytes), including the Firmware name, but I couldn't detect it because it is ASCII (7-bits), and I had parsed the firmware name incorrectly.

I was refactoring the parser's firmware string decoder to use your logic for strings, when I realized what I had done.

I have the pull request ready to patch...

zfields:
I discovered it on the ESP8266, because your node client parses the firmware name. I didn't see it before, because the analog data scrolls so fast and I never scrolled back up to the top.

soundanalogous:
I2C also appears to be broken in 2.5.5

soundanalogous:
Your latest update fixes the I2C issue as well. Looks like in 2.5.5 all sysex data was being parsed as strings.

PR #364, opened 2017-03-18
codec bug

zfields:
#363

### DIFF
--- a/Firmata.cpp
+++ b/Firmata.cpp
@@ -50,7 +50,7 @@ systemCallbackFunction FirmataClass::currentSystemResetCallback = (systemCallbac
  */
 void FirmataClass::sendValueAsTwo7bitBytes(int value)
 {
-  marshaller.transformByteStreamToMessageBytes(sizeof(value), reinterpret_cast<uint8_t *>(&value), sizeof(value));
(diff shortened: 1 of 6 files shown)


zfields:
I have tested this on all my platforms, and for backward compatibility with 2.5.4.

soundanalogous:
I didn't run https://github.com/firmata/arduino/pull/359 through my tests or I would have caught this. Lesson learned. It's also time we added a string parser test to [firmata_test.ino](https://github.com/firmata/arduino/blob/master/test/firmata_test/firmata_test.ino).

soundanalogous:
See page 29 of [this pdf](http://breakoutjs.com/examples/schematics.pdf). This is the circuit I use to test several features at once (adc, pwm, servo, digital on digital pins, digital I/O on analog pins, I2C read, I2C write). I should really get this made into a shield at some point so other contributors can test with the same setup (and then port the demo app to their respective client libraries). I use Breakout.js for this particular test. The code [is here](https://github.com/soundanalogous/Breakout/blob/master/examples/boards/uno.html) (in case you wanted to create an equivalent example for remote-wiring).

soundanalogous:
Ran through my tests and all is good.

PR #365, opened 2017-03-18
add tests for sending and receiving strings

soundanalogous:
@zfields FYI


zfields:
Nice! I will be running this every time! :+1:

PR #370, opened 2017-04-19
Add support for Arduino MKRFox1200

sandeepmistry:
This adds support for the new Arduino MKRFox1200 board.

Please note that this board is only available in Europe for now, see: https://blog.arduino.cc/2017/04/18/introducing-the-arduino-mkrfox1200/

@agdl has briefly tested this change on with a MKRFox1200 and Johny-Five.

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -247,7 +247,7 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #elif defined(ARDUINO_SAMD_MKR1000)
 #define TOTAL_ANALOG_PINS       7
 #define TOTAL_PINS              22 // 8 digital + 3 spi + 2 i2c + 2 uart + 7 analog
-#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 21) && !IS_PIN_SERIAL(p))
+#define IS_PIN_DIGITAL(p)       ((p) >= 0 && (p) <= 21)
 #define IS_PIN_ANALOG(p)        ((p) >= 15 && (p) < 15 + TOTAL_ANALOG_PINS)
 #define IS_PIN_PWM(p)           digitalPinHasPWM(p)
 #define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS) // deprecated since v2.4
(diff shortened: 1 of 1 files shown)


Boards.h
```diff
@@ -276,6 +276,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
+// Arduino MKRFox1200
+#elif defined(ARDUINO_SAMD_MKRFox1200)
+#define TOTAL_ANALOG_PINS       7
+#define TOTAL_PINS              33 // 8 digital + 3 spi + 2 i2c + 2 uart + 7 analog + 3 usb + 1 aref + 5 sd + 1 bottom pad + 1 battery adc
+#define IS_PIN_DIGITAL(p)       ((((p) >= 0 && (p) <= 21)) && !IS_PIN_SERIAL(p))
```
soundanalogous: Are pins 13 and 14 also used for the primary serial transport (`Serial` instead of `Serial1`, `Serial1`, etc) or are they otherwise special UART pins that cannot be used as GPIO?

Boards.h
```diff
@@ -276,6 +276,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define IS_PIN_ANALOG(p)        (((p) >= 15 && (p) < 15 + TOTAL_ANALOG_PINS) || (p) == 32)
```
soundanalogous: Are pins 22 - 31 unused?

soundanalogous: or otherwise not able to be set as analog or digital I/O?

Boards.h
sandeepmistry: Pins 13 and 14 are used for `Serial1`. `Serial` is a native USB port.

We might have to correct the MKR1000 and MKRZero info. that this was based on ...

Boards.h
sandeepmistry: They are not broken out or useable from the user's perspective:

* 23 - 24 are for USB
* 25 is for AREF
* 26 - 31 are connected to the built-in SigFox module (SPI + reset + power on)

https://github.com/arduino/ArduinoCore-samd/blob/master/variants/mkrfox1200/variant.cpp#L105-L137

Boards.h
soundanalogous: But is there any reason these pins cannot be used for digital I/O? It has not been an issue with other boards (since `Serial1.begin()` is only called if the user sends a SERIAL_CONFIG message). Only pins tied to `Serial` cannot be used for digital I/O on other boards.

soundanalogous: Ah I see now, I must have overlooked that `!IS_PIN_SERIAL(p)` was used for the MKR1000 and MKRZero as well. Yeah I don't think that is necessary.

sandeepmistry: Ok, i'll push a commit to correct this for all of those boards.

sandeepmistry: Removed this check in https://github.com/firmata/arduino/pull/370/commits/d4a9ba96706bb1a979049bc63c0bf824e390a218.

soundanalogous:
LGTM. Thanks!!

PR #375, opened 2017-07-09
added server mode to StandardFirmataEthernet

MJPees:
server mode implemented.
For server mode remote_ip, remote_host and remote_port should not be
defined.
Instead the server_port has to be defined.

### DIFF
--- a/examples/StandardFirmataEthernet/StandardFirmataEthernet.ino
+++ b/examples/StandardFirmataEthernet/StandardFirmataEthernet.ino
@@ -26,8 +26,8 @@
 /*
   README
 
-  StandardFirmataEthernet is a TCP client implementation. You will need a Firmata client library
-  with a network transport that can act as a TCP server in order to establish a connection between
+  StandardFirmataEthernet is a TCP client/server implementation. You will need a Firmata client library
+  with a network transport that can act as a TCP server or client in order to establish a connection between
   StandardFirmataEthernet and the Firmata client application.
 
   To use StandardFirmataEthernet you will need to have one of the following
@@ -68,6 +68,7 @@
 // follow the instructions in ethernetConfig.h to configure your particular hardware
 #include "ethernetConfig.h"
 #include "utility/EthernetClientStream.h"
+#include "utility/EthernetServerStream.h"
 
 /*
  * Uncomment the following include to enable interfacing with Serial devices via hardware or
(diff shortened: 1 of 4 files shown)


examples/StandardFirmataEthernet/ethernetConfig.h
```diff
@@ -55,7 +56,9 @@ YunClient client;
 
 // STEP 3 [REQUIRED]
 // Replace with the port that your server is listening on
+// for client mode comment out remote_port / for server mode comment out server_port
 #define remote_port 3030
```
soundanalogous: This could be simplified by renaming `remote_port` to `network_port` and then using that for both client and server applications.

soundanalogous: The update the comment to:
```c++
// Replace with the port that your client or server is listening on.
```

examples/StandardFirmataEthernet/ethernetConfig.h
```diff
@@ -3,8 +3,9 @@
  *
  * You must configure your particular hardware. Follow the steps below.
  *
- * Currently StandardFirmataEthernet is configured as a TCP client. An
- * option to configure as a server may be added in the future.
+ * Currently StandardFirmataEthernet is configured as a TCP client.
```
soundanalogous: Please change this text to:
"By default, StandardFirmataEthernet is configured as a TCP client. To configure as a TCP server, see STEP 2."

Then update STEP 2 to read as follows:
```c++
// STEP 2 [REQUIRED for all boards and shields]
// TCP Client configuration:
// To configure your board as a TCP client, set the IP address of the server you want to connect to.
// TCP Server configuration:
// To configure your board as a TCP server, comment out the following line and also ensure that
// remote_host is also commented out.
```

examples/StandardFirmataEthernet/StandardFirmataEthernet.ino
```diff
@@ -939,7 +944,11 @@ void loop()
 #if !defined local_ip && !defined YUN_ETHERNET
   // only necessary when using DHCP, ensures local IP is updated appropriately if it changes
   if (Ethernet.maintain()) {
+    #if defined remote_ip || defined remote_host
     stream.maintain(Ethernet.localIP());
+    #else
+    stream.maintain();
```
soundanalogous: I don't think you need to call stream.maintain() here since it is already called on each iteration of loop when firmata.available is called. However, you may need to copy [the maintain(IPAddress local) method](https://github.com/firmata/arduino/blob/master/utility/EthernetClientStream.h#L101) from EthernetClientStream to EthernetServerStream and then leave this block of code as it originally was. You should test this however to ensure there are not side effects.

MJPees:
I think the parameter client is useless at initialization of EthernetServerStream.h => removed it.
Thanks so far!

examples/StandardFirmataEthernet/StandardFirmataEthernet.ino
```diff
@@ -103,17 +104,25 @@
...
+EthernetClientStream stream(client, local_ip, IPAddress(0, 0, 0, 0), remote_host, network_port );
 #else
-EthernetClientStream stream(client, IPAddress(0, 0, 0, 0), IPAddress(0, 0, 0, 0), remote_host, remote_port);
+EthernetClientStream stream(client, IPAddress(0, 0, 0, 0), IPAddress(0, 0, 0, 0), remote_host, network_port);
+#endif
+#endif
+
+#if !defined remote_ip && !defined remote_host
+#ifdef local_ip
+EthernetServerStream stream(client, local_ip, network_port);
```
soundanalogous: Please remove the `client` parameter here and on line 125. Then make sure StandardFirmataEthernet compiles when configured for TCP server.

MJPees:
Tested on Arduino Ethernet and Arduino Mega/ Ethernet shield

soundanalogous:
I finally got around to thoroughly testing this today and it's working without issue. However, one addition would be helpful. Please add the following 3 lines at the end of the `initTransport()` function in StandardFirmataEthernet.ino (just after the `DEBUG_PRINT("connecting...")` line, but add a space in-between):

```c++
DEBUG_PRINT("IP Address: ");
IPAddress ip = Ethernet.localIP();
DEBUG_PRINTLN(ip);
```

That will print the local IP address when debugging is enabled, which is helpful for the user to know what the board's IP address is when configured as a TCP server.

After that, please squash all commits to a single commit. Then this PR will be ready to merge. Thanks for all the work so far!

PR #377, opened 2017-07-29
Single-Read Functionality. 

Molorius:
This adds the functionality to read a digital or analog pin, as opposed to the current method of reporting. This is useful in non-threaded environments, such as GNU Octave, when constant data is either a hindrance or unnecessary. 

The command uses the 0xA0 block to read an analog pin and 0xB0 to read a digital port. I did this to keep it in line with the analog and digital reporting commands, but it can be moved to a different block if this is not ideal. 

The structure of the new commands are identical to the reporting commands. It still requires a second byte, but the value is not checked. 

I have only updated the StandardFirmata.ino to accommodate this change right now, but I can update the others if this merge is accepted. 

Since this adds new functionality, I also changed the version from 2.5.6 to 2.6.0.


Molorius:
I am going to change the command so that either:
- The commands stay on the 0xA0 and 0xB0 blocks, with no additional byte, or
- I put the commands on 0xF2 and 0xF3, with the following byte specifying the analog pin or digital port.

Thoughts on either?

soundanalogous:
@Molorius, please start a thread here to propose this feature: https://github.com/firmata/protocol.

PR #378, opened 2017-08-01
Add support for STM32 MCU based boards

fpistm:
Hi,

This PR add the support of STM32 MCU based board.
More information here: [STM32 cores enabled in Arduino IDE](https://community.st.com/community/stm32-community/blog/2017/02/02/stm32-cores-enabled-in-arduino-ide)
Latest release of the [Arduino_Core_STM32](https://github.com/stm32duino/Arduino_Core_STM32) is functional with Firmata. The modified lib with ARDUINO_ARCH_STM32 is currently released with the core and could be removed when this PR will be merged.
Core is provided as a package thanks the Arduino Boards manager, see the [wiki](https://github.com/stm32duino/wiki/wiki/Getting-Started)

Thanks in advance

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -780,6 +780,26 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p)
 #define DEFAULT_PWM_RESOLUTION  10
 
+// STM32 based boards
+#elif defined(ARDUINO_ARCH_STM32)
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              NUM_DIGITAL_PINS
+#define TOTAL_PORTS             MAX_NB_PORT
+#define VERSION_BLINK_PIN       LED_BUILTIN
+// PIN_SERIALY_RX/TX defined in the variant.h
+#define IS_PIN_DIGITAL(p)       (digitalPinIsValid(p) && !pinIsSerial(p))
+#define IS_PIN_ANALOG(p)        ((p >= A0) && (p < AEND) && !pinIsSerial(p))
+#define IS_PIN_PWM(p)           (IS_PIN_DIGITAL(p) && digitalPinHasPWM(p))
+#define IS_PIN_SERVO(p)         IS_PIN_DIGITAL(p)
+#define IS_PIN_I2C(p)           (IS_PIN_DIGITAL(p) && digitalPinHasI2C(p))
+#define IS_PIN_SPI(p)           (IS_PIN_DIGITAL(p) && digitalPinHasSPI(p))
+#define IS_PIN_INTERRUPT(p)     (IS_PIN_DIGITAL(p) && (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)))
+#define IS_PIN_SERIAL(p)        (digitalPinHasSerial(p) && !pinIsSerial(p))
+#define PIN_TO_DIGITAL(p)       (p)
+#define PIN_TO_ANALOG(p)        (p-A0)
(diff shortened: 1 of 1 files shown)


Boards.h
```diff
@@ -780,6 +780,26 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+#define VERSION_BLINK_PIN       LED_BUILTIN
+// PIN_SERIALY_RX/TX defined in the variant.h
+#define IS_PIN_DIGITAL(p)       (digitalPinIsValid(p) && !pinIsSerial(p))
+#define IS_PIN_ANALOG(p)        ((p >= A0) && (p < AEND) && !pinIsSerial(p))
+#define IS_PIN_PWM(p)           (IS_PIN_DIGITAL(p) && digitalPinHasPWM(p))
+#define IS_PIN_SERVO(p)         IS_PIN_DIGITAL(p)
+#define IS_PIN_I2C(p)           (IS_PIN_DIGITAL(p) && digitalPinHasI2C(p))
+#define IS_PIN_SPI(p)           (IS_PIN_DIGITAL(p) && digitalPinHasSPI(p))
+#define IS_PIN_INTERRUPT(p)     (IS_PIN_DIGITAL(p) && (digitalPinToInterrupt(p) > NOT_AN_INTERRUPT)))
+#define IS_PIN_SERIAL(p)        (digitalPinHasSerial(p))
```
soundanalogous: If any UART pins share the main USB/UART programming connection, please exclude those pins here (whatever maps to `Serial` instead of `Serial1`, `Serial1`, etc).

fpistm: You mean something like that?
`#define IS_PIN_SERIAL(p)        (digitalPinHasSerial(p) && !pinIsSerial(p) )`

soundanalogous: Yes that is correct as defined [here](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/pins_arduino.h#L77).

Also is seems strange that there are Firmata-specific macros defined in `Arduino_Core_STM32`. It's convenient for sure, but ideally board cores shouldn't need to know anything specific about Firmata. What is really missing is a well defined HAL that exposes board capabilities across the entire Arduino ecosystem. This is one area that Arduino only loosely defined so you see a lot of variance in macros defined across all of the various boards defined for the Arduino ecosystem. If such a system were in place and properly maintained, there would be no need for the Firmata Boards.h file. Some day perhaps (wishful thinking).

fpistm: Agreed with you.
In fact the core know anything about Firmata. It only give a macro that Firmata could use for STM32 arch and define the PIN_SERIAL_RX && PIN_SERIAL_TX which are different for each STM32 based board.

Note about [this line](https://github.com/stm32duino/Arduino_Core_STM32/blob/master/cores/arduino/pins_arduino.h#L77 ): Due to the pin mapping implementation and allow the most possibilities for each pins/boards some pins are duplicated to handle Analog input.
So this one is used to ensure to not break the Serial link btw Firmata client and Firmware.

About a well defined HAL that exposes board capabilities... Sounds great but seems hard to align all the Arduino ecosystem.
That's why I tried to use proper macro definition/name to avoid any further change request in Firmata/arduino.

soundanalogous:
Can you squash these 2 commits? Then this will be ready to merge.

fpistm:
Done. Thanks in advance.

fpistm:
Hi Jeff,
Just for my information, when the next release is planned ?
Thanks in advance.
Fred

soundanalogous:
There are a couple of changes I want to make to StandardFirmataEthernet before I push the next release. Sometime this weekend would be the earliest, but it could end up being another week or two.

PR #379, opened 2017-08-09
add Adafruit nrf52 boards support

hathach:
This Boards.h modification add support for Adafruit nrf52 board series.
https://www.adafruit.com/product/3406

Unfortunately our ble API is a bit different than the currently one used in examples/StandardFirmataBLE.ino
Thus we have to make separate sketch in our example repo.
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/libraries/Bluefruit52Lib/examples/Peripheral/StandardFirmataBLE/StandardFirmataBLE.ino

But this patch will allow user to compile without us including a separated fork

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -801,6 +801,24 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p)
 #define DEFAULT_PWM_RESOLUTION  PWM_RESOLUTION
 
+// Adafruit Bluefruit nRF52 boards
+#elif defined(ARDUINO_NRF52_ADAFRUIT)
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              32
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < TOTAL_PINS)
+#define IS_PIN_ANALOG(p)        ((p) == PIN_A0 || (p) == PIN_A1 || (p) == PIN_A2  || (p) == PIN_A3 || \
+                                 (p) == PIN_A4 || (p) == PIN_A5 || (p) == PIN_A6  || (p) == PIN_A7)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         IS_PIN_DIGITAL(p)
+#define IS_PIN_I2C(p)           ((p) == PIN_WIRE_SDA || (p) == PIN_WIRE_SCL)
+#define IS_PIN_SPI(p)           ((p) == SS || (p)== MOSI || (p) == MISO || (p == SCK))
+#define PIN_TO_DIGITAL(p)       (p)
+#define PIN_TO_ANALOG(p)        ( ((p) == PIN_A0) ? 0 : ((p) == PIN_A1) ? 1 : ((p) == PIN_A2) ? 2 : ((p) == PIN_A3) ? 3 : \
+                                  ((p) == PIN_A4) ? 4 : ((p) == PIN_A5) ? 5 : ((p) == PIN_A6) ? 6 : ((p) == PIN_A7) ? 7 : (127))
+#define PIN_TO_PWM(p)           (p)
+#define PIN_TO_SERVO(p)         (p)
+
 // anything else
 #else
 #error "Please edit Boards.h with a hardware abstraction for this board"


Boards.h
```diff
@@ -780,6 +780,25 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         (p)
 #define DEFAULT_PWM_RESOLUTION  10
 
+// Adafruit nrf52 boards
+#elif defined(ARDUINO_NRF52_ADAFRUIT)
+#define TOTAL_ANALOG_PINS       NUM_ANALOG_INPUTS
+#define TOTAL_PINS              32
+#define VERSION_BLINK_PIN       LED_BUILTIN
+#define IS_PIN_DIGITAL(p)       ((p) >= 2 && (p) < TOTAL_PINS)
+#define IS_PIN_ANALOG(p)        ((p) == PIN_A0 || (p) == PIN_A1 || (p) == PIN_A2  || (p) == PIN_A3 || \
```
soundanalogous: Are these not consecutive pin numbers?

Boards.h
```diff
@@ -780,6 +780,25 @@ writePort(port, value, bitmask):  Write an 8 bit port.
...
+                                 (p) == PIN_A4 || (p) == PIN_A5 || (p) == PIN_A6  || (p) == PIN_A7)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         IS_PIN_DIGITAL(p)
+#define IS_PIN_I2C(p)           ((p) == PIN_WIRE_SDA || (p) == PIN_WIRE_SCL)
+#define IS_PIN_SPI(p)           ((p) == SS || (p)== MOSI || (p) == MISO || (p == SCK))
+#define IS_PIN_SERIAL(p)        ((p) == PIN_SERIAL_RX || (p) == PIN_SERIAL_TX)
```
soundanalogous: The primary Serial port `Serial` (as opposed to `Serial1`, `Serial2`, etc) should not be reported.

Boards.h
hathach: kind of. They are 2-5, 28-31. But we have 2 boards with different order of An.
- feather nrf52:  2 (A0), 3, 4, 5, 28, 29, 30, 31 ( A7 )
- metro nrf52 :   30 (A0), 31, 2, 3, 4 ,5 (A5)

There could be more in the future. But we will update the macro

soundanalogous:
This may be easier to maintain if you break a lot of the BLE functionality out into a BLEStream wrapper class. That way when StandardFirmata is updated it's less work to keep this BLE version up to date.

Boards.h
soundanalogous: I see... what you have makes sense then.

hathach:
We have stream bleuart
https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/libraries/Bluefruit52Lib/src/services/BLEUart.h

However since our own BLE libraries's API is not compatible with Sandeep's BLEPeripheral library. Could we add an separated file let's say BluefruitStream.h ?

Boards.h
soundanalogous: I'm going to add support someday to use `Serial` when the a non-Serial transport such as BLE, WiFi or Ethernet is used, but I just haven't gotten there yet.

soundanalogous:
If you could make BluefruitStream work in such a way that no major changes (especially no #ifdefs) would need to be added to StandardFirmataBLE, then that would be ideal. If you take that route then you can also add a section in bleConfig.h following the existing patterns for currently supported boards.

hathach:
Thanks for your suggestion, we will give it a shot. Will push update when ready.

hathach:
hi @soundanalogous, 

Follow your suggestion, we have added the BluefruitStream.h and update the bleconfig.h, it work beautifully with your StandardFirmataBLE.ino without any modification. Everything works well except for an issue with **analogRead()** that we don't know how to resolve. Let's say, pin A4 in our board is 28  
- in your example sketch, it is `analogRead(4)`
- however our core implementation require it to be `analogRead(A4)` where A4 = 28, which is essentially analogRead(28). That's why we added an ANALOG_TO_PIN() in our own example https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/libraries/Bluefruit52Lib/examples/Peripheral/StandardFirmataBLE/StandardFirmataBLE.ino#L49

We have an internal discussion whether we should change the parameter from A4 (28) to simple number 4 to meet your example. However, since it is already released and customers are using it. Changing the API will cause user confusion and lots of support question etc ... Do you have any suggestion, we are leaning toward an folk now although we are really want to get it run with the official Firmata.

soundanalogous:
This is an issue I've been seeing more frequently as new Arduino cores role out. Historically Arduino has supported passing either the analog channel or pin number to `analogRead`. You can see it implemented [here in wiring_analog.c](https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/wiring_analog.c#L42) for AVR boards. Could you do something similar [here in the Adafruit_nRF52_Arduino core](https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/wiring_analog_nRF52.c#L127)?

I think it would also be helpful at some point for Arduino to express a stronger set of rules to be met when new cores are implemented.

hathach:
Implement mapping between analog channel to pin number is easy. We had an internal discussion on this, but decided not to do so because the code is already released, changing API behavior could confuse user.

Since we are on the run to release the next version (delayed so far), we may include the folk as part of the release if there is no walk-around for analogRead() any soon. This is unfortunate, but we could always come back later when there is something new.

Thank you very much for your time and suggestion so far. We appreciate it.

examples/StandardFirmataBLE/bleConfig.h
```diff
@@ -106,6 +106,15 @@ BLEStream stream;
 // BLEStream stream(BLE_REQ, BLE_RDY, BLE_RST);
 // #endif
 
+/*
+ * Adafruit Bluefruit nRF52 Boards
+ */
+#ifdef ARDUINO_NRF52_ADAFRUIT
+#include "utility/BluefruitStream.h"
+BLEStream stream;
+#endif
```
soundanalogous: You could define the `ANALOG_TO_PIN` macro here if that solves the `analogRead` issue.

soundanalogous: If so add a comment in the code documenting that it's a work around.

hathach: ANALOG_TO_PIN macro needs to be used by the example sketch though

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/libraries/Bluefruit52Lib/examples/Peripheral/StandardFirmataBLE/StandardFirmataBLE.ino#L475

https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/develop/libraries/Bluefruit52Lib/examples/Peripheral/StandardFirmataBLE/StandardFirmataBLE.ino#L887

Just quick question does analogPinToChannel macro is Arduino standard way to detect this conversion ? Maybe firmata can use that, core like ours need to define this to get it work in the board #ifdef ?

https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/wiring_analog.c#L42

hathach: ah, I kind of guess what you mean. Though the wiring_analog.c does not include the bleconfig.h there is no way for it to know if the macro existed.

soundanalogous: I see I overlooked the way you are using ANALOG_TO_PIN so my suggestion here would not work. I could try to resolve this on the Firmata side by allowing either a channel or pin number but it may be a while before I have time to work on a fix.

hathach: yeah, we will release the code the folk just now to make user happy. There is no breaking change for the pull request, we will come back to this issue later then. Thank you very much for your suggestion so far.

soundanalogous: To clarify, do you want to hold off on this pull request until a proper solution is in place or do you still need the Boards.h changes?

hathach: hi @soundanalogous 
If you agree to merge the Boards.h now, it would be super great for us. Then we don't need to include a folk of Firmata. We could point user to our own example now and says we are working on solution to integrate to your stock example. Please merge the Boards.h if possible :+1:

soundanalogous: Can you open a separate PR with just the Boards.h changes. I'll then leave this PR open until the analogRead issues are resolved.

hathach: Sorry for late response, this is the pull request that contains only Boards.h for adafruit nrf52 boards series. Thank you very much
https://github.com/firmata/arduino/pull/381

PR #383, opened 2017-08-17
Fix I2C config parameter interpretation

zfields:
The current implementation pulls bytes out of the sysex buffer, regardless of whether or not the user specified such bytes.

### DIFF
--- a/examples/StandardFirmata/StandardFirmata.ino
+++ b/examples/StandardFirmata/StandardFirmata.ino
@@ -586,7 +586,7 @@ void sysexCallback(byte command, byte argc, byte *argv)
     case I2C_CONFIG:
       delayTime = (argv[0] + (argv[1] << 7));
 
-      if (delayTime > 0) {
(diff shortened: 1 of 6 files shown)


fpistm:
Access to argv[1] before test argc is risked? no ?

zafields:
In this circumstance, there is no risk, because `argv` is a statically allocated buffer:
https://github.com/firmata/arduino/blob/e0592e7c34a2b87fd164c83ad236fb074d8f94fe/Firmata.h#L114

You can see it being supplied to the callback:
https://github.com/firmata/arduino/blob/e0592e7c34a2b87fd164c83ad236fb074d8f94fe/FirmataParser.cpp#L455

You are right, in saying it would be best practice to do so, and to protect against the buffering scheme being changed. For now I want a surgical fix, before I implement a much broader change to the firmware.

rwaldron:
This change would need to be made across all variants.

soundanalogous:
lgtm!

PR #393, opened 2018-03-28
Add support for Adafruit Feather M0 Bluefruit LE to StandardFirmataBLE

cstawarz:
These changes add support for the [Adafruit Feather M0 Bluefruit LE](https://www.adafruit.com/product/2995) board to StandardFirmataBLE.  They should also work with other Bluefruit LE boards/modules that communicate with the nRF51822 via SPI, although I've only tested the Feather M0.  The changes require a few [small modifications](https://github.com/cstawarz/Adafruit_BluefruitLE_nRF51/tree/firmata_fixes) to the Adafruit BluefruitLE nRF51 package (pull request: adafruit/Adafruit_BluefruitLE_nRF51#43).

I also moved the configuration of the BLE connection and flush intervals to bleConfig.h and updated the connection interval values for the Arduino 101 (which, as of Intel Curie Boards package v2.0.2, must be given in units of milliseconds, not 1.25ms).

I tested the changes on both an Arduino 101 and a Feather M0 Bluefruit LE.

### DIFF
--- a/examples/StandardFirmataBLE/StandardFirmataBLE.ino
+++ b/examples/StandardFirmataBLE/StandardFirmataBLE.ino
@@ -768,6 +768,13 @@ void setup()
   Firmata.attach(START_SYSEX, sysexCallback);
   Firmata.attach(SYSTEM_RESET, systemResetCallback);
 
+  stream.setLocalName(FIRMATA_BLE_LOCAL_NAME);
+
+  // set the BLE connection interval - this is the fastest interval you can read inputs
+  stream.setConnectionInterval(FIRMATA_BLE_MIN_INTERVAL, FIRMATA_BLE_MAX_INTERVAL);
+  // set how often the BLE TX buffer is flushed (if not full)
+  stream.setFlushInterval(FIRMATA_BLE_TXBUFFER_FLUSH_INTERVAL);
+
 #ifdef IS_IGNORE_BLE_PINS
   for (byte i = 0; i < TOTAL_PINS; i++) {
     if (IS_IGNORE_BLE_PINS(i)) {
--- a/examples/StandardFirmataBLE/StandardFirmataBLE.ino
+++ b/examples/StandardFirmataBLE/StandardFirmataBLE.ino
@@ -768,7 +768,7 @@ void setup()
   Firmata.attach(START_SYSEX, sysexCallback);
   Firmata.attach(SYSTEM_RESET, systemResetCallback);
 
-#ifdef BLE_REQ
+#ifdef IS_IGNORE_BLE_PINS
   for (byte i = 0; i < TOTAL_PINS; i++) {
     if (IS_IGNORE_BLE_PINS(i)) {
       Firmata.setPinMode(i, PIN_MODE_IGNORE);
--- a/examples/StandardFirmataBLE/StandardFirmataBLE.ino
+++ b/examples/StandardFirmataBLE/StandardFirmataBLE.ino
@@ -56,10 +56,6 @@
 // the minimum interval for sampling analog input
 #define MINIMUM_SAMPLING_INTERVAL   1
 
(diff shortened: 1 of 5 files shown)


soundanalogous:
Thanks! I'll try to find time to look at this more closely over the weekend.

examples/StandardFirmataBLE/StandardFirmataBLE.ino
```diff
@@ -772,14 +768,7 @@ void setup()
   Firmata.attach(START_SYSEX, sysexCallback);
   Firmata.attach(SYSTEM_RESET, systemResetCallback);
 
-  stream.setLocalName(FIRMATA_BLE_LOCAL_NAME);
```
soundanalogous: I can live with this change, but the reason I used these setters initially was to avoid tightly coupling defines in bleConfig.h with the BLE stream classes. The only reason your change here works is all the code in BLEStream is in the header file, otherwise BLEStream would have been precompiled and the defines in bleConfig.h would not be available due to the weird way the Arduino liker works.

soundanalogous:
Overall LGTM, I had one comment about removal of some setters.

soundanalogous:
I think you may also need to add the Adafruit Feather M0 Bluefruit LE to Boards.h unless it's pin compatible with an existing M0 board already defined (my hunch is that it is not).

examples/StandardFirmataBLE/bleConfig.h
```diff
@@ -5,20 +5,37 @@
...
+/*
+ * Arduino 101
+ *
+ * Make sure you have the Intel Curie Boards package v2.0.2 or higher installed via the Arduino
+ * Boards Manager.
+ *
+ * Test script: https://gist.github.com/soundanalogous/927360b797574ed50e27
+ */
+#ifdef _VARIANT_ARDUINO_101_X_
+// After conversion to units of 1.25ms, both values must be between
```
soundanalogous: We can remove all the comments here since the user doesn't need to be concerned with the conversion anymore. I'd just update the comments on lines 34 and 35 to say `// 8ms` and `// 30ms` like you did for the Bluefruit configuration.

examples/StandardFirmataBLE/StandardFirmataBLE.ino
cstawarz: Good point.  I was trying to simplify things, but I didn't consider this issue.  I'll add the setters back.

examples/StandardFirmataBLE/bleConfig.h
cstawarz: I'd say it's still useful to know that, under the hood, the values are converted to multiples of 1.25ms -- if only to explain why using the value 8 actually gets you an interval of 7.5ms.

cstawarz:
> I think you may also need to add the Adafruit Feather M0 Bluefruit LE to Boards.h unless it's pin compatible with an existing M0 board already defined

The Feather M0 masquerades as an Arduino Zero.  Its [variant.h](https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/feather_m0/variant.h) even defines `_VARIANT_ARDUINO_ZERO_`.  This appears warranted, as the pin mappings look to be identical to the Zero's.

The only difference I see is that the Feather has two more analog inputs (A6 and A7).  However, those inputs don't follow the digital_pin=analog_pin+14 numbering that A0-A5 do, and the analogInputToDigitalPin macro defined in variant.h doesn't handle them.  Personally, I'm OK with ignoring them.

examples/StandardFirmataBLE/bleConfig.h
soundanalogous: ok sgtm

cstawarz:
This should be good to go now.  Thanks!

soundanalogous:
Ok, I'll confirm on my Arduino 101 just to make sure there were no unexpected regressions.

Issue #407, opened 2018-10-18
Arduino MKR1010 support


eccoscreen:
Hi will you guys be adding support for the MKR WiFi1010? I would try to edit Boards.h but Windows won't actually let me edit that file.

Thanks!

soundanalogous:
This will enable a sketch to compile for this board, but I don't think it will enable WiFi support. For that you'd also have to edit [wifiConfig.h](https://github.com/firmata/arduino/blob/master/examples/StandardFirmataWiFi/wifiConfig.h)

PR #413, opened 2018-12-26
#407: Add MKR1010 support

dowlingw:
To address #407

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -342,6 +342,22 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_PWM(p)           PIN_TO_DIGITAL(p)
 #define PIN_TO_SERVO(p)         (p) // deprecated since v2.4
 
+// Arduino MKR WiFi 1010
+#elif defined(ARDUINO_SAMD_MKRWIFI1010)
+#define TOTAL_ANALOG_PINS       7
+#define TOTAL_PINS              33
+#define IS_PIN_DIGITAL(p)       (((p) >= 0 && (p) <= 14))
+#define IS_PIN_ANALOG(p)        ((p) >= 15 && (p) < 15 + TOTAL_ANALOG_PINS)
+#define IS_PIN_PWM(p)           digitalPinHasPWM(p)
+#define IS_PIN_SERVO(p)         (IS_PIN_DIGITAL(p) && (p) < MAX_SERVOS) // deprecated since v2.4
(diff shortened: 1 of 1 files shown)


soundanalogous:
This will enable a sketch to compile for this board, but I don't think it will enable WiFi support. For that you'd also have to edit wifiConfig.h

soundanalogous:
Hummm, looking at the MKR 1010, I see it's ESP32 based. I'm not sure how it's architected so I'm not sure exactly what it will take to get it to work over WiFi with Firmata. It's probably more complicated than updating wifiConfig.h.

dowlingw:
Thanks for that, I've spotted another omission also (total number of pins, though I'm not 100% sure how that's supposed to be counted).

I had a look at the WiFi config and it looks like there's already some support for ESP32 based WiFi, my read was that the `platform.h` provides some `#define`s that `wificonfig.h` will see and automatically include support for.

I'll double check and have a look - super new to Arduino in general so not 100% confident.

soundanalogous:
`TOTAL_PINS` is a bit tricky. It's actually the highest addressable digital pin number. Early on this was always just count of digital pins + count of analog pins (because all analog pins could also be configured as digital pins). As new boards have been introduced over the years, this is not always so simple anymore. So if the highest digital pin number (per the Arduino numbering scheme which likely differs from the pin numbers on the microcontroller itself) is say "60" then TOTAL_PINS should be set to 60 and IS_PIN_DIGITAL needs to filter out any pins in the range of 0 to 60 that are not actually configurable as digital input or output pins.

soundanalogous:
Regarding WiFi support. If the MKR 1010 uses the WiFi101.h library, then changing 34 in wifiConfig.to `#if (defined(ARDUINO_SAMD_MKR1000) || defined(ARDUINO_SAMD_MKRWIFI1010)) && !defined(WIFI_101)` may work. Worth a try at least. If you need a Firmata client that supports the WiFi transport, see [firmata.js](https://github.com/firmata/firmata.js/) and [johnny-five.js](https://github.com/rwaldron/johnny-five). I have some examples for using both of those clients with WiFi [here](https://github.com/soundanalogous/firmata-client-examples) (see the `network` folder in both the `firmata-js` and `johnny-five` folders. Sorry there's not a ton of documentation on using Firmata over WiFi and it can be a bit testy.

dowlingw:
Sorry guys, life things happened - closing as #418 sailed on through :)

PR #418, opened 2019-04-04
Add support for Arduino MKR WiFi 1010

gdsports:
@dowlingw @eccoscreen 
The board uses the U-Blox WiFi NINA chip which has an ESP32 inside running
NINA firmware. The SAMD21 is connected to the U-Blox chip via SPI. The
WiFININA library is used to communicate between the U-Blox chip and the
SAMD21. Other boards use the same U-Blox chip such as the MKR Vidor 4000
and the Uno WiFi Rev2. Additional changes will be required to support
those boards. See #407

### DIFF
--- a/Boards.h
+++ b/Boards.h
@@ -261,8 +261,8 @@ writePort(port, value, bitmask):  Write an 8 bit port.
 #define PIN_TO_SERVO(p)         ((p) - 2)
 
 
-// Arduino/Genuino MKR1000
-#elif defined(ARDUINO_SAMD_MKR1000)
+// Arduino/Genuino MKR1000 or MKR1010
+#elif defined(ARDUINO_SAMD_MKR1000)	|| defined(ARDUINO_SAMD_MKRWIFI1010)
 #define TOTAL_ANALOG_PINS       7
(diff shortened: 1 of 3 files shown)


soundanalogous:
Which manual tests have you run against this configuration?

gdsports:
So far only this code. LED blinks and floating A0 returns random garbage.

```
var Firmata = require("firmata").Board;
var EtherPortClient = require("etherport-client").EtherPortClient;

var board = new Firmata(new EtherPortClient({
	host: "192.168.1.116",
	port: 3030
}));

board.on("ready", function() {
  console.log("READY!");
  console.log(
    board.firmware.name + "-" +
    board.firmware.version.major + "." +
    board.firmware.version.minor
  );

  var state = 1;
  var lastVal = 0;

  this.pinMode(6, this.MODES.OUTPUT);

  setInterval(function() {
    this.digitalWrite(6, (state ^= 1));
  }.bind(this), 500);

  this.analogRead(0, function(value) {
    if (value != lastVal) {
      console.log(value);
    }
  });

});
```

soundanalogous:
Ok looks good. Thanks!

PR #422, opened 2019-05-17
Set Bluefruit LE advertising interval

cstawarz:
By default, the BLE firmware on the Adafruit Feather M0 Bluefruit LE will switch from fast advertising (20ms interval) to low-power advertising (417.5ms interval) after 30 seconds.  I found that, after the switch to low-power advertising, connection attempts from macOS devices often failed (despite the fact that such a switch is explicitly condoned in [Apple's documentation](https://developer.apple.com/library/archive/qa/qa1931/_index.html)).

These changes alter the default configuration of Bluefruit LE devices to always use the initial, 20ms advertising interval.  They also introduce a new preprocessor macro, `FIRMATA_BLE_ADVERTISING_INTERVAL`, which can be set by other device configurations.  Users who want to stick with the Feather M0's default advertising behavior can simply comment out the macro definition in `bleConfig.h`.

### DIFF
--- a/examples/StandardFirmataBLE/StandardFirmataBLE.ino
+++ b/examples/StandardFirmataBLE/StandardFirmataBLE.ino
@@ -770,6 +770,10 @@ void setup()
 
   stream.setLocalName(FIRMATA_BLE_LOCAL_NAME);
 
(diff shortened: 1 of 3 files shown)


soundanalogous:
Thanks!

PR #431, opened 2019-10-20
Add support for SPRESENSE board

baggio63446333:
Add support for SONY SPRESENSE board

Reference:
https://developer.sony.com/develop/spresense/


soundanalogous:
Thank you!

PR #438, opened 2019-12-09
Fix build issue with arm none eabi gcc v9.2.1

### LINKED ISSUE #437: Build issue after arm none eabi gcc update
Hi,

I've updated the arm none eabi gcc toolchain used by the STM32 core to the v9.2.1.
I've built all Arduino built-in example without any issue except those for `Firmata`, see the error hereafter:

```C++
Firmata/Firmata.h:131:17: error: friend declaration of 'void encodeByteStream(size_t, uint8_t*, size_t) const' specifies default arguments and isn't a definition [-fpermissive]
  131 |     friend void FirmataMarshaller::encodeByteStream (size_t bytec, uint8_t * bytev, size_t max_bytes = 0) const;
      |                 ^~~~~~~~~~~~~~~~~
```

As specified by the compiler output, error comes from the default argument here:
https://github.com/firmata/arduino/blob/1ccb2c00e2b28f4ce7486d645d14fee71810c9fd/Firmata.h#L131

As it is also specified here:
https://github.com/firmata/arduino/blob/1ccb2c00e2b28f4ce7486d645d14fee71810c9fd/FirmataMarshaller.h#L67

I guess it could be removed from the friend declaration as already declare in `FirmataMarshaller.h`

Ref:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2000/n1263.html

fpistm:
Firmata/Firmata.h:131:17: error: friend declaration of 'void encodeByteStream(size_t, uint8_t*, size_t) const' specifies default arguments and isn't a definition [-fpermissive]
  131 |     friend void FirmataMarshaller::encodeByteStream (size_t bytec, uint8_t * bytev, size_t max_bytes = 0) const;
      |                 ^~~~~~~~~~~~~~~~~

Ref:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2000/n1263.html

Fixes #437

### DIFF
--- a/Firmata.h
+++ b/Firmata.h
@@ -128,7 +128,7 @@ class FirmataClass
 
     /* private methods ------------------------------ */
     void strobeBlinkPin(byte pin, int count, int onInterval, int offInterval);
-    friend void FirmataMarshaller::encodeByteStream (size_t bytec, uint8_t * bytev, size_t max_bytes = 0) const;
+    friend void FirmataMarshaller::encodeByteStream (size_t bytec, uint8_t * bytev, size_t max_bytes) const;
 
     /* callback functions */
     static callbackFunction currentAnalogCallback;


zfields:
@soundanalogous Do we know if the default value will still propogate from the declaration in `FirmataMarshaller.h`?

https://github.com/firmata/arduino/blob/master/FirmataMarshaller.h#L67

I think your fix is the right way to handle this, but I want to be sure we aren't breaking the API.

soundanalogous:
> Do we know if the default value will still propagate from the declaration in FirmataMarshaller.h?

Not exactly sure. There are 2 instances where a 3rd parameter isn't passed to `encodeByteStream` if that is what you're asking. You can find them [Here](https://github.com/firmata/arduino/blob/master/FirmataMarshaller.cpp#L329) and [here](https://github.com/firmata/arduino/blob/master/FirmataMarshaller.cpp#L396) in the FirmataMarshaller implementation.

soundanalogous:
Tested with a bunch of examples from Breakout.js, didn't encounter any issues.

fpistm:
Thanks

PR #447, opened 2020-04-05
Host stream write buffer for message oriented transmits

jnsbyr:
Firmata never really moved on from its serial MIDI roots to frame oriented host connections like Ethernet. It took me a while to figure it out but the problem is easy to verify with a network sniffer like tcpdump: every data byte is currently transferred separately. With TCP/IP the minimum overhead is factor 40, with WiFi even higher. No wonder the W5100 gets burning hot, switch LEDs blink like wild and the host CPU load is higher than expected e.g. when using Serial Firmata to transfer 64 byte messages every 100 millis. This example is not everyone's use case but still there is something simple that can be done to remedy this problem for all use cases: adding a write buffer to the host stream implementations like EthernetStream and WiFiStream that gets flushed with every call to Stream.maintain() at the end of each Firmata main loop. 

Using a write buffer has the following advantages:
- write buffer writes are faster than direct Ethernet transmits
- less SPI transfers needed to network controllers (e.g. W5100)
- Firmata loop frequency / sampling rate / performance increases
- network load decreases: example 3 data bytes for a digital message, before: 3 TCP frames with >120 network bytes, after: 1 TCP frame with ~43 network bytes
- host CPU load decreases with the number of received TCP frames
- total energy consumption decreases (Firmata device, network, host)


jnsbyr:
This feature can be enabled and configured in EthernetClientStream.h, EthernetServerStream.h and WiFiStream.h respectively by

`#define WRITE_BUFFER_SIZE 40`

It is enabled by default. The default size may be increased depending on the use case, e.g. for Serial Firmata and long messages. 

From my point of view it would be preferable to handle this feature generically for all host stream implementations e.g. by introducing a FirmataStream base class, but this would require changes to several classes like Firmata and FirmataMarshaller.

PR #450, opened 2020-04-19
Return only as many bytes as were received in I2C transfers.

pgrawehr:
Same as https://github.com/firmata/ConfigurableFirmata/pull/94. 

Tested for StandardFirmata.


soundanalogous:
Thanks again!

PR #455, opened 2020-05-17
 Simplify all StandardFirmata examples

d-a-v:
A new file is added `utility/ExampleStandardFirmataCommon.h` which is 99% of `StandardFirmata.ino` (`diff -u` is given below because github won't show it).

All other StandardFirmata examples were/are based on this file. The goal is to simplify them all as well and obtain a common source code for all "Standard Firmata" examples.
This file can also can be included from user projects without having to copy the full code, which in turn may be improved upstream sometimes (which is always better).

```
--- StandardFirmata.ino.orig	2020-05-17 18:21:37.772320715 +0200
+++ ../../utility/ExampleStandardFirmataCommon.h	2020-05-17 18:21:42.180322855 +0200
@@ -11,7 +11,7 @@
   Copyright (C) 2006-2008 Hans-Christoph Steiner.  All rights reserved.
   Copyright (C) 2010-2011 Paul Stoffregen.  All rights reserved.
   Copyright (C) 2009 Shigeru Kobayashi.  All rights reserved.
-  Copyright (C) 2009-2016 Jeff Hoefs.  All rights reserved.
+  Copyright (C) 2009-2017 Jeff Hoefs.  All rights reserved.
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
@@ -23,17 +23,23 @@
   Last updated August 17th, 2017
 */
 
-#include <Servo.h>
-#include <Wire.h>
-#include <Firmata.h>
-
-#define I2C_WRITE                   B00000000
-#define I2C_READ                    B00001000
-#define I2C_READ_CONTINUOUSLY       B00010000
-#define I2C_STOP_READING            B00011000
-#define I2C_READ_WRITE_MODE_MASK    B00011000
-#define I2C_10BIT_ADDRESS_MODE_MASK B00100000
-#define I2C_END_TX_MASK             B01000000
+//#include <Firmata.h> <-- sketch must include this file
+
+#ifndef SERVO
+#define SERVO Servo
+#endif
+
+#ifndef ANALOGWRITE
+#define ANALOGWRITE_PIN_MODE_PWM analogWrite
+#endif
+
+#define I2C_WRITE                   0x00 //B00000000
+#define I2C_READ                    0x08 //B00001000
+#define I2C_READ_CONTINUOUSLY       0x10 //B00010000
+#define I2C_STOP_READING            0x18 //B00011000
+#define I2C_READ_WRITE_MODE_MASK    0x18 //B00011000
[...]

PR #456, opened 2020-06-01
Standard Firmata through websockets

d-a-v:
Websockets allow a WiFi-enabled Firmata device to be natively controlled by a web browser.

An example is provided:
- esp8266 arduino sketch (using two external libraries, see in comments)
- a poor javascript proof of concept allowing to blink the arduino builtin led without further setup
  (sorry not to be an html5 expert)

This PR is used for https://github.com/bromagosa/Snap4Arduino/issues/297
This PR relies on #455 - For an easy review, [diff against #455](https://github.com/d-a-v/FirmatArduino/compare/commoncode...d-a-v:ws?expand=1) is provided


soundanalogous:
Can you update this PR to exclude all the changes that involve: `#include <utility/ExampleStandardFirmataCommon.h>`. I'd prefer it only includes the changes relevant to adding websocket support.

d-a-v:
@soundanalogous This PR is updated

soundanalogous:
Had some time to think more about this proposal. I definitely like the idea of websocket support for Firmata and have pondered it several times over the past few years. I even created a Firmata client library that relied on websockets (via a serial to websocket server).

Historically we have not included architecture specific example with Firmata. I think this proposal is off to a good start, but what I would like to see is support (or at least a plan) for additional boards such as some of the newer Arduino WiFi boards (such as the Nano 33 IoT and MKR WiFi 1010) which should have enough RAM and high enough clock speeds to perform well with a websocket implementation.

d-a-v:
> but what I would like to see is support (or at least a plan) for additional boards such as some of the newer Arduino WiFi boards

Do you think of a particular PAL¹ for websockets, or should we make a new one with specific names ?
From what I understand of what is necessary in this websockets use case, it is not more than what's in the event function:
- connected event
- disconnected event
- receive event (binary? text? both? data, len)

and a write function (binary? text? both? data, len)

Underlying protocol consideration (ping, ...) would left to the PAL implementer. 

¹ Protocol Abstraction Layer

soundanalogous:
Yeah, something like we used for [BLEStream](https://github.com/firmata/arduino/blob/master/utility/BLEStream.h) as an abstraction layer that works with different BLE libraries.

d-a-v:
That's a stream without events.
This is partly done thanks to the pipedstreams used in this PR's example.
I will isolate and hide details and offer a Stream as websocket API.
I like the idea very much, it looks like a [previous attempt](https://github.com/d-a-v/libNoWebsocket) I made for websockets in posix world.

Issue #457, opened 2020-07-11
Issue when compiling StandardFirmataWiFi for MKR WiFi 1010 board


fredbouchard:
## Version
**Firmata:** latest from `master`
**Arduino IDE:** `1.8.13`
**Board:** `Arduino MKRWIFI1010 board` 
**OS version:** macOS Catalina `10.15.4`

## What am I trying to achieve?

I'm trying to compile the `StandardFirmataWiFi.ino`or the `StandardFirmata.ino` with the MKR 1010 board.

## Issue

When compiling I'm getting the following error.

```cpp
/Users/{Redacted}/Documents/Arduino/libraries/Firmata/utility/SerialFirmata.cpp: In member function 'void SerialFirmata::checkSerial()':
/Users/{Redacted}/Documents/Arduino/libraries/Firmata/utility/SerialFirmata.cpp:364:56: error: 'SERIAL_RX_BUFFER_SIZE' was not declared in this scope
           if (!((bytesToRead <= 0 && bytesAvailable >= SERIAL_RX_BUFFER_SIZE/2)
                                                        ^~~~~~~~~~~~~~~~~~~~~
/Users/{Redacted}/Documents/Arduino/libraries/Firmata/utility/SerialFirmata.cpp:364:56: note: suggested alternative: 'SERIAL_BUFFER_SIZE'
           if (!((bytesToRead <= 0 && bytesAvailable >= SERIAL_RX_BUFFER_SIZE/2)
                                                        ^~~~~~~~~~~~~~~~~~~~~
                                                        SERIAL_BUFFER_SIZE
```

soundanalogous:
Fixed. Thanks for the report!

soundanalogous:
Actually my fix is only temporary. If [FIRMATA_SERIAL_RX_DELAY](https://github.com/firmata/arduino/blob/master/utility/SerialFirmata.h#L56) is uncommented, the issue will still occur. The real problem is that UART_TX_FIFO_SIZE does not appear to be defined for Arduino M0 board packages, so SERIAL_RX_BUFFER_SIZE never gets defined, leading to the compile time issue pointed out here. I'll look to @jnsbyr to provide a fix. It may be an issue for other architectures supported by Firmata. I've only tested for M0 and AVR so far. There was not issue with AVR boards.

jnsbyr:
This is a platform and hardware dependent issue I am unable to resolve by
myself because my use cases are limited to AVR and ESP8266

[...]

PR #458, opened 2020-07-12
Fix compile error for M0 boards

soundanalogous:
FIRMATA_SERIAL_RX_DELAY being defined by default resulted [SERIAL_RX_BUFFER_SIZE not being defined](https://github.com/firmata/arduino/blob/e385ece47cfe361f36714fa7f20552058d0229ea/utility/SerialFirmata.cpp#L364) for Arduino M0 boards, resulting in a compile error. Temporary fix is to keep FIRMATA_SERIAL_RX_DELAY commented out by default.
@jnsbyr FYI.


soundanalogous:
Fixed issue #457

PR #464, opened 2020-08-19
Import Arduino.h rather than Stream.h

soundanalogous:
Per [this note](https://github.com/arduino/ArduinoCore-API/blob/master/api/deprecated/Stream.h#L19) in the Steam.h header file. I'm still trying to understand what led to this change. Apparently Arduino is using a [common core library](https://github.com/arduino/ArduinoCore-API) and if you try to compile against an board from one of the 4 cores referenced in the [readme](https://github.com/arduino/ArduinoCore-API/blob/master/README.md), then you'll get an error regarding Stream.h not being found. Changing to Arduino.h per the note in the Stream.h header seemed to have resolved the issue. Not sure yet if there are other potential issues to be aware of per these changes.

### DIFF
--- a/FirmataMarshaller.h
+++ b/FirmataMarshaller.h
@@ -22,7 +22,7 @@
   #include <stdint.h>
 #endif
 
-#include <Stream.h>
+#include <Arduino.h>
 
 namespace firmata {
 


soundanalogous:
To repro the error I encountered, install the Arduino megaAVR boards package from the Boards Manager, then try to compile StandardFirmata (latest version in the master branch).

zfields:
Without the full Arduino dependency, then `FirmataMarshaller` can be ported to other C++ environments beyond the Arduino (e.g. Linux, RemoteWiring, etc.). By including `Arduino.h`, we bring in all the dependencies of Arduino.

My original intention for `FirmataMarshaller` was to generate the byte protocol. Theoretically, it should not require a dependency on Arduino to arrange bytes in to/out of our protocol.

Instead of solving it this way, let me see if I can drop the `#include "Stream.h"` statement altogether, by simply using a pointer in the API, and forward declaring the class.

soundanalogous:
Sounds good.

zfields:
Okay, I did a little research this weekend. Unfortunately, my quick fix will not appease the compiler, or address the larger problem of deprecation.

I propose that we leverage an interface to `Stream`, and/or use dependency injection. Unfortunately, doing so will break the `FirmataMarshaller` public API. Although, I don't really think we make guarantees about it, but I do believe people are using it.

Not only can we not support the megaAVR boards, but Arduino is forcing our hand with the deprecation of `Stream`. We definitely need to do something. @soundanalogous What are your feelings on breaking the FirmataMarshaller public API, specifically the `.begin(Stream &s)` method?

PR #474, opened 2020-11-29
add nano-33-iot to Boards.h

soundanalogous:


soundanalogous:
Untested other than verified StandardFirmata compiles. I don't have this board.

PR #485, opened 2021-10-30
Add support for ArduinoBLE devices to StandardFirmataBLE

cstawarz:
These changes extend StandardFirmataBLE to work with boards supported by the [ArduinoBLE](https://www.arduino.cc/en/Reference/ArduinoBLE) library.  I've only tested on the [Nano 33 IoT](https://docs.arduino.cc/hardware/nano-33-iot), but any other ArduinoBLE-compatible board should work, too.

### DIFF
--- a/utility/ArduinoBLE_UART_Stream.cpp
+++ b/utility/ArduinoBLE_UART_Stream.cpp
@@ -1,3 +1,4 @@
 /*
- * Implementation is in ArduinoBLE_UART_Stream.h to avoid linker issues.
+ * Implementation is in ArduinoBLE_UART_Stream.h to avoid making ArduinoBLE a
+ * build-time dependency for all projects that use the Firmata library.
  */
--- a/utility/ArduinoBLE_UART_Stream.cpp
+++ b/utility/ArduinoBLE_UART_Stream.cpp
@@ -1,0 +1,3 @@
+/*
+ * Implementation is in ArduinoBLE_UART_Stream.h to avoid linker issues.
+ */
--- a/examples/StandardFirmataBLE/bleConfig.h
+++ b/examples/StandardFirmataBLE/bleConfig.h
@@ -1,15 +1,15 @@
 /*==================================================================================================
  * BLE CONFIGURATION
  *
- * If you are using an Arduino 101, you do not need to make any changes to this file (unless you
- * need a unique ble local name (see below). If you are using another supported BLE board or shield,
- * follow the instructions for the specific board or shield below.
- *
- * Make sure you have the Intel Curie Boards package v2.0.2 or higher installed via the Arduino
- * Boards Manager.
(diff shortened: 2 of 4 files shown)


utility/ArduinoBLE_UART_Stream.cpp
```diff
@@ -0,0 +1,3 @@
+/*
+ * Implementation is in ArduinoBLE_UART_Stream.h to avoid linker issues.
```
pgrawehr, line 2: Are you sure this is necessary? What kind of linker errors do you see when you do it the "normal" way?

cstawarz: The issue is that, if you move the class implementation to the cpp file, the Arduino IDE will *always* compile it, even if the current project doesn't require it and/or the current board is not compatible with it.

For example, if the method implementations are in the cpp file (i.e. the "normal" way), and you try to build StandardFirmata for an Arduino Uno, compilation fails because ArduinoBLE is missing:
```
ArduinoBLE_UART_Stream.cpp:5:0:
/Users/cstawarz/Documents/Arduino/libraries/Firmata/utility/ArduinoBLE_UART_Stream.h:11:10: fatal error: ArduinoBLE.h: No such file or directory
 #include <ArduinoBLE.h>
          ^~~~~~~~~~~~~~
compilation terminated.
```
If you then install ArduinoBLE, even though you obviously don't need it, you get more build failures:
```
/Users/cstawarz/Documents/Arduino/libraries/ArduinoBLE/src/utility/HCIUartTransport.cpp:33:2: error: #error "Unsupported board selected!"
 #error "Unsupported board selected!"
  ^~~~~
/Users/cstawarz/Documents/Arduino/libraries/ArduinoBLE/src/utility/HCIUartTransport.cpp:99:40: error: 'SerialHCI' was not declared in this scope
 HCIUartTransportClass HCIUartTransport(SerialHCI, 912600);
                                        ^~~~~~~~~
```

The "avoid linker issues" bit was just a lazy copy from BLEStream.cpp.  Should I change it to say "avoid an unnecessary dependency on the ArduinoBLE library"?

pgrawehr: Ok, that makes sense, even though it's a bit ugly. Yes, please change that comment, then it will be fine for me.

cstawarz: Done!  I also changed the comment in BluefruitLE_SPI_Stream.cpp to clarify why it, too, is empty.

pgrawehr (approved):
Thanks for adding this. 

Would you mind updating ConfigurableFirmata as well? In the long run, StandardFirmata will probably be deprecated, because it's advantages don't justify maintaining two code bases any more.

cstawarz:
Sorry, but I don't use ConfigurableFirmata.  Someone who does (and wants these changes there) should port this over.

PR #495, opened 2022-09-15
Added Teensy 4.1 pins to Boards.h

ktgilliam:
Added Teensy 4.1 pins to Boards.h.  #define TEENSY40 or TEENSY41 to select. Defaults to TEENSY40 with warning if neither is defined.


pgrawehr:
Thanks for the update. I do recommend to use ConfigurableFirmata for new developments, though.

pgrawehr (approved):

PR #507, opened 2023-08-10
Improve defines for Teensy 4.0 and Teensy 4.1

PaulStoffregen:
This pull request allows Boards.h to automatically detect Teensy 4.0 vs Teensy 4.1 without requiring the user to manually define anything.  No actual code is changed, only the `#if defined()` checks are updated to the names which are automatically defined within the Arduino environment.


pgrawehr (approved):
Thanks for this cleanup

PR #510, opened 2023-09-03
Documentation: Add reference to the Firmata Linux Kernel module

logicog:
Adds a reference to the (experimental) Linux Kernel Module firmata_mod, which allows users to use standard Linux userspace APIs/tools to access GPIO, I2C or SPI devices made available on an Arduino attached to a Linux host PC via a serial interface and running the Firmata firmware. In addition, existing Linux Kernel modules for devices using I2C/SPI can also be associated with devices on the Arduino. For example, it is possible to use sensors on the
Arduino board with the standard iio or hwmon Linux drivers. An example would
be an SHT21 humity/heat sensor that can be out-of-the box made accessible e.g
on the KDE desktop without any additional code to be written.


pgrawehr (approved):

logicog:
Thanks a lot!

TCB13:
Thank you very much for taking the time to get this up and running.

PR #517, opened 2024-09-09
Fix compilation for UNOR4

pgrawehr:


zfields:
If I understand the fix correctly, it appears that all digital pins will be marked as PWM capable, but this doesn't appear to be true based on the pinout for the MINIMA or WIFI.

- [MINIMA Pinout](https://docs.arduino.cc/resources/pinouts/ABX00080-full-pinout.pdf)
- [WIFI Pinout](https://docs.arduino.cc/resources/pinouts/ABX00087-full-pinout.pdf)

pgrawehr:
I need to check that again, but that's what the mapping was doing already, except that it used a macro that basically caused a circular dependency. Something is broken there, and I currently don't know why it ever worked.

pgrawehr:
@zfields Now this makes more sense. It seems the previously conflicting definitions where now resolved and it just compiles without these definitions. I don't have the board, so I can only test the compilation, not whether it works.

zfields:
I don't have one either. I say let's merge it for now, and see if anyone complains.

zfields (approved):

pgrawehr:
Ok, fine with me. I'll fix it for ConfigurableFirmata as well, has the exactly same issue.

